PR #1, opened 2017-06-08
linode: specify a default for stackscriptjson

dwfreed:
Prevents the API from getting bad values.


Dorthu:
Thanks!  This will make it into tomorrow's release, 1.0.3b1

Dorthu:
This was deployed in https://github.com/linode/linode-cli/releases/tag/v1.0.3b1.  Thank you again!

Issue #22, opened 2018-03-10
Cannot remove domain records with duplicate names.


StefanLiebenberg:
When I create multiple A records for the same subdomain, I cannot remove any of them again through with cli.

```sh
$ linode-cli domain record-create -y A -n "*" -R "192.168.0.1" domain.com
$ linode-cli domain record-create -y A -n "*" -R "192.168.0.2" domain.com
$ linode-cli domain record-create -y A -n "*" -R "192.168.0.3" domain.com
```

```sh
$ linode-cli domain record-delete domain.com A "*"
Ambiguous criteria - found 3 records instead of 1
```

Reading through the code, it seems that if I do not target only one machine to use the cli in this way, but I can only use the address or the label. While I could use the address in my example, this won't always work as I might have other labels pointing to the address.

Ideally, I would like a --all option to bypass the check to ensure that I've only selected one machine.

PR #23, opened 2018-03-10
Allows CLI to delete multiple domains

StefanLiebenberg:
This fixes https://github.com/linode/linode-cli/issues/22.

Please excuse my bad python.


Dorthu:
The CLI has changed drastically since this was submitted, and I'm not sure it makes sense anymore.  Sorry it never made it in.

PR #25, opened 2018-05-04
configure accept more ssh key types

camper42:
accept more key types
https://github.com/linode/linode-api-python/blob/master/linode/common.py#L26


Dorthu:
The CLI was rewritten since this PR was opened, and this is no longer relevant.  Sorry I missed it 😞

Issue #34, opened 2018-07-01
make install failed on Ubuntu 17.10


ginkgomzd:
using master or latest 2.x tag
```
mzd☯ artichoke :>sudo make install SPEC=https://developers.linode.com/openapi.yaml
python linodecli bake https://developers.linode.com/openapi.yaml --skip-config
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/usr/local/share/linode-cli/linodecli/__main__.py", line 3, in <module>
    from linodecli import main
ImportError: No module named linodecli
Makefile:20: fallo en las instrucciones para el objetivo 'common'
make: *** [common] Error 1
```

ginkgomzd:
I tried this, which looks like an error to me (I don't know any python).

```
mzd☯ artichoke :>git diff
diff --git a/Makefile b/Makefile
index 45c3c41..a3bb0d0 100644
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,6 @@ build: common
        $(PYCMD) setup.py bdist_wheel --universal
 
 common:
-       python linodecli bake ${SPEC} --skip-config
-       python3 linodecli bake ${SPEC} --skip-config
+       $(PYCMD) linodecli bake ${SPEC} --skip-config
        cp data-2 linodecli/
        cp data-3 linodecli/

```
... but there was no change in the error, other than

ginkgomzd:
so, I got brave, and I googled, and I changed __main__.py

```
root@localhost:/usr/local/sbin/linode-cli# git diff
diff --git a/linodecli/__main__.py b/linodecli/__main__.py
index 98f3fa9..47153f7 100644
--- a/linodecli/__main__.py
+++ b/linodecli/__main__.py
@@ -1,5 +1,5 @@
 from __future__ import absolute_import
 
-from linodecli import main
+from __init__ import main
 
 main()
```

Which appears to have worked, but then hit another error:
```
  File "linodecli/__init__.py", line 6, in <module>
    import pkg_resources
ImportError: No module named pkg_resources
```

Dorthu:
I think I got it - can you please pull master and verify it fixed so I can close this?

ginkgomzd:
It does look like the first error I reported is fixed, but it is again getting stuck on the second, for "pkg_resources"

[...]

PR #37, opened 2018-07-03
fixes issue #34

ginkgomzd:
I see no reason that make install shouldn't be a one-stop shopping experience.


Dorthu:
This is awesome - thank you!  I have a follow up PR that builds on this a little more, I would appreciate your feedback on it if you have time to take a look.

PR #38, opened 2018-07-04
Couple of ideas to further improve the builder

Dorthu:
I really like what @ginkgomzd with the build process - I'm not going to
lie, this is my first Makefile, and I appreciate the cleanup.  I had a
few ideas looking it over:

If we're going to build a package during the install process,
why not install the package that we just built?  I think that works
better anyway - once you test the package that was installed, you can
confidently distribute it knowing that it is the same code you just ran.

I also  added a clean step to delete all of the generated files
(linodecli/data-* and linode-cli.sh along with the contents of dist/) to
make sure everything's fresh with each build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,6 @@ check-prerequisites:
 
 .PHONY: clean
 clean:
-	rm linodecli/data-*
-	rm linode-cli.sh
+	rm -f linodecli/data-*
+	rm -f linode-cli.sh
 	rm -f dist/*
--- a/Makefile
+++ b/Makefile
@@ -7,23 +7,22 @@ SPEC ?= https://developers.linode.com/openapi.yaml
 
 ifeq ($(PYTHON), 3)
 	PYCMD=python3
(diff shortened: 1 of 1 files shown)


ginkgomzd:
Looks good. 
Only thing is I don't know if this is a problem for installing alternate builds. Since `install` will automatically re-build, I don't know if that is taking too much control away from developers. 

If the final step of `build` is `touch build`, then build will only run when being called directly (or when needed). 
https://www.gnu.org/software/make/manual/html_node/Empty-Targets.html#Empty-Targets

Dorthu:
That's fair, but old `install` behavior was a `$(PYCMD) setup.py install`, which does a build on its own (and ignores the package that `build` produces).  I think this is better - at least you're installing from a package that is like what is actually distributed.  Also, if the spec changes on developers.linode.com, nothing on the filesystem would've changed, and a `make install` would unexpectedly not rebuild the package.

PR #41, opened 2018-07-23
POC - Add BATS tests

patthiel:
* This PR adds some initial support for BATS (bash automated testing system) test files.
* Added bats-assert and bats-support sub-modules
* Added some initial test cases for the /linodes endpoint
* Added a `common.bash` file that will contain common helper functions used in other bats test files.

# Testing

1. Install BATS (on mac, `brew install bats-core` or see : https://github.com/bats-core/bats-core/#installation) for non-mac installs
2. checkout this pr
3. `git submodule init && git submodule update`
4. Run the example test from the base of the repo: `bats test/linodes`
⚠️ The test suite will delete all Linodes on the account that runs the tests, even those that existed before the tests were run ⚠️ 
5. Observe, _*tests*_!

# To Do

* Document how to install bats and run the tests
* Update tests to not assume a `.linode-cli` file exists
* Make root password random, not hard coded


test/linodes/linodes.bats
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bats
+
+load '../test_helper/bats-support/load'
+load '../test_helper/bats-assert/load'
+load '../common'
+
+@test "it should create linodes with a label" {
+    run linode-cli linodes create --type g6-standard-2 --region us-east --image $test_image --label cli-1 --root_pass $random_pass --text --delimiter "," --no-headers
+    assert_output --regexp ".*,cli-1,us-east,g6-standard-2,$test_image,.*"
```
Dorthu: You could use a `--format 'label,region,type,image'` to remove the need for the `.*`s in this regex

test/linodes/linodes.bats
```diff
@@ -0,0 +1,36 @@
...
+    run linode-cli linodes view "$linode_id" --text --delimiter "," --no-headers
+    assert_output --regexp "$linode_id,cli-1,us-east,g6-standard-2,$test_image,.*"
+}
+
+@test "it should create a linode with the minimum required props" {
+    run linode-cli linodes create --type g6-standard-2 --region us-east --root_pass $random_pass
+    [ "$status" -eq 0 ]
+}
+
+# ONLY WORKS BECAUSE I HAVE A .linode-cli
```
Dorthu: To be fair, it's also using that file for the token it's running the tests with, but you are correct in saying that if you hadn't chosen a default image this would fail (adding a `--image $test_image` or similar would remove this condition)

test/linodes/linodes.bats
```diff
@@ -0,0 +1,36 @@
...
+@test "it should fail to create a linode without a root_pass" {
+    run linode-cli linodes create --type g6-standard-2 --region us-east
+    [ "$status" -eq 1 ]
```
Dorthu: if you included a `--text --no-headers` you could test for output like this (in addition to testing for the status):

```
Request failed: 400
root_pass	root_pass is required
```

test/linodes/linodes.bats
patthiel: ah, good idea!

test/linodes/linodes.bats
patthiel: so, i tested this on prod last night, and realized that this is permissible on prod, but not staging? slightly confused, do i need to modify this in order for it to work across all envs?

patthiel:
@Dorthu I believe I have addressed all feedback. Let me know if you would like any additional changes. I will plan to work on more smoke tests this week. 😺

PR #44, opened 2018-08-13
BATS tests Dockerfile

patthiel:
* Adds Dockerfile to run BATS tests
* Adds dockerignore file so that builds/dist files are ignored when creating the docker image

* This PR was based off the domains test branch, I will rebase once that PR is merged. The main files of concern here are `Dockerfile-bats` `.dockerignore` and `README.rst`

## Testing

***Please remember these TESTS **WILL DELETE ALL RESOURCES** (linodes/volumes/domains/nodebalancers/etc/etc) associated with your linode account!***

* Follow the instructions for running via docker in the README

cc. @Dorthu

### DIFF
--- a/Dockerfile-bats
+++ b/Dockerfile-bats
@@ -12,6 +12,7 @@ ENV PATH="/usr/local/bin:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin
 
 WORKDIR /src
 
+# Build and Install Linode CLI
 RUN git clone https://github.com/linode/linode-cli.git \
     && cd linode-cli \
     && git submodule init \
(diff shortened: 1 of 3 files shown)


Dockerfile-bats
```diff
@@ -0,0 +1,33 @@
...
+COPY . .
+
+# Build and Install Linode CLI
+RUN git submodule init \
+    && git submodule update \
+    && make build SPEC=${SPEC} \
+    && make install SPEC=${SPEC} \
+    && cd dist \
+    && pip install --user $(ls) \
+    && echo -n "[DEFAULT]\ntoken = ${TOKEN}" > /root/.linode-cli
```
Dorthu: This config file will work, but a real config file would have keys for `image`, `type`, and `region`

Issue #51, opened 2018-09-27
Add --suppress-warnings flag


Dorthu:
Some users have experienced problems with scripts when the "Upgrade" warning appeared following a new API release.  This can be especially frustrating if the CLI update is not immediately available (they're usually close, but not timed exactly the same).  A flag to suppress this warning should help alleviate these issues.

PR #55, opened 2018-10-06
feat(#51): Added `--suppress-warnings` flag to ignore newer API versions

alxbl:
Does exactly as the PR says.

If there are more warnings that this flag should suppress, it should be a matter of adding a check for `self.suppress_warnings` before the warnings are output.

If you provide me with a list of such code locations or strings, I can quickly add them to this PR.

Cheers,


alxbl:
Oops, I just realized I probably forgot to `git add __init__.py`. I saw you added the switch already though, sorry for that.

PR #59, opened 2018-10-12
Add Volumes tests to test Dockerfile command

patthiel:
* Adds the volumes tests directory to the BATS dockerfile start command 
* Updates the test runner to look for `.bats` files in the test directory
* updates the assertion for the list volumes test to accept `creating` or `active` response from the API, because volume creation is quick! apparently.

### DIFF
--- a/Dockerfile-bats
+++ b/Dockerfile-bats
@@ -43,4 +43,4 @@ RUN git clone https://github.com/bats-core/bats-core.git \
     && ./install.sh /usr/local
 
 
-CMD bats $(ls test/*/*.bats)
+CMD bats $(ls /src/linode-cli/test/*/*.bats)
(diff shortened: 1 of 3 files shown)


Dockerfile-bats
```diff
@@ -33,7 +33,7 @@ RUN curl -o /src/linode-cli/openapi.yaml ${SPEC} \
     && make install SPEC=cli-tests.yaml \
     && cd dist \
     && pip install --user $(ls) \
-    && echo -n "[DEFAULT]\ntoken = ${TOKEN}" > /root/.linode-cli
+    && echo -n "[DEFAULT]\ntoken = ${TOKEN}\n" > /root/.linode-cli
```
Dorthu: You can use the `LINODE_CLI_TOKEN` environment variable now, if that's easier 😉

PR #63, opened 2018-10-17
Updated URL to public OpenAPI spec

Dorthu:
This changed upstream, so it needed to be updated.  Builds without this
change will fail.


patthiel:
can you update the spec url in here as well: https://github.com/linode/linode-cli/blob/master/Dockerfile-bats

PR #68, opened 2018-11-07
Added Markdown output format

### LINKED ISSUE #67: Provide a markdown output format
Markdown could be handled in a similar way to the current `--text` tabbed delimited format.

Tabs would be replaced with ` | `, and the header line would be followed by a line with `---` field for each field name (also delimited by `|`).

Dorthu:
Closes #67

As requested, prints out output in a github-styled markdown format.

Example usage: `linode-cli regions list --markdown` outputs this:

| id | country |
|---|---|
| us-central | us |
| us-west | us |
| us-southeast | us |
| us-east | us |
| eu-west | uk |
| ap-south | sg |
| eu-central | de |
| ap-northeast | jp |
| ap-northeast-1a | jp |

### DIFF
--- a/linodecli/output.py
+++ b/linodecli/output.py
@@ -159,7 +159,7 @@ class OutputHandler:
             content=data
         else:
             for model in data:
-                content.append([attr.render_value(model) for attr in columns])
+                content.append([attr.render_value(model, colorize=False) for attr in columns])
 
         if header:
             print('| ' + ' | '.join([str(c) for c in header]) + ' |')
--- a/linodecli/output.py
+++ b/linodecli/output.py
@@ -10,6 +10,7 @@ class OutputMode(Enum):
     table=1
     delimited=2
     json=3
+    markdown=4
 
 
 class OutputHandler:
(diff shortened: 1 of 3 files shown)


patthiel:
@Dorthu  Can you disable colorized text when selecting markdown output?

linodecli/output.py
```diff
@@ -144,3 +147,23 @@ def _select_json_elements(self, keys, json):
...
+        else:
+            for model in data:
+                content.append([attr.render_value(model, colorize=False) for attr in columns])
+
+        if header:
+            print('| ' + ' | '.join([str(c) for c in header]) + ' |')
+            print('|---' * len(header) + '|')
+
+        for row in content:
+            print('| ' + ' | '.join([str(c) for c in row]) + ' |')
```
displague: Should be concerned with escaping (`\_`, `\|`)?

Underscore is definitely possible in common fields.

Breaking characters, like pipes and newlines, could cause edge-case hardships.

I wouldn't hold this PR up on either of those.

Dorthu:
What's the temperature on this?

displague:
I had to manually markdown'ify the cli output just the other day to update https://gist.github.com/displague/3c1728fdf4ff2bacf6960a9b8c6ea73f

I'm still interested in this, and it looks good to me.

PR #69, opened 2018-11-10
Plugins

Dorthu:
Plugins are python modules included with the CLI that are not generated
from the OpenAPI spec.  At this time, plugins must be accepted in the
mainline CLI repo to be installed.  See README.md for more information
on plugin support.


Dorthu:
@patthiel @asauber I believe this is ready to merge.  Please give it a look and let me know what you think

patthiel:
I had trouble running the `statuspage` plugin. The instructions to `pip install beautifulsoup` didn't work because python 3 is my default version. I resolved this by running `pip install beautifulsoup4` which i believe will install the latest for either python 2 or 3.

PR #72, opened 2018-12-02
[WIP] ssh-into plugin

Dorthu:
The intended goal of this plugin is to allow the following command to
work: `linode-cli ssh-into my-linode`.  This plugin will use the CLI to
query the API for information about the linode by label and execute an
ssh command into that Linode if found.

Remaining work:
- [ ] Should this accept IDs in addition to/instead of labels?
- [x] Make this "work" (or fail with reasonable errors) on Windows
- [ ] Confirm the ssh approach is secure and sane

### DIFF
--- a/linodecli/__init__.py
+++ b/linodecli/__init__.py
@@ -180,14 +180,16 @@ def main():
         cli.bake_completions()
 
     # check for plugin invocation
-    if parsed.command not in cli.ops and parsed.command in plugins.available:
+    plugin_command = parsed.command.replace('-', '_')
+
+    if parsed.command not in cli.ops and plugin_command in plugins.available:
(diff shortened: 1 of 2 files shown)


ilsken:
One other suggestion would be to hack in support for `user@host` by doing the following:

```
# after parsing args
label = parsed.label.split('@', 2).pop()
# use label for the API calls and search etc instead of parsed.label

# execute this
# ssh will see username@label and use username but will still use our Hostname override 
subprocess.check_call(['ssh', '-oHostname={}'.format(public_ip)]', parsed.label] + parsed.rest)
```

displague:
`lish-into` support should be easy to tack onto this.

Dorthu:
This is being closed in favor of #121

PR #74, opened 2018-12-03
ARB-771 - Nodebalancers BATS Smoke Tests

patthiel:
*Fair **warning**, running the automated CLI tests will remove all entities on your account. Please update your CLI configuration profile to a test user before running!*

* Added tests for nodebalancer creation, creating a nodebalancer configuration profile, adding a node to the nodebalancer, editing an existing profile, adding an additional profile, removing a profile
* Updated the CLI tests dockerfile to suppress warnings about API versions by default
* Added removeAll utility method (will potentially refactor existing tests to use this instead of specific "removeLinodes" "removeDomains" etc methods.

## To Test:

* Assumes you have the bats framework installed locally (`brew install bats-core`)
```bash
 bats test/nodebalancers/nodebalancers.bats
```

* and plz read the test code

### DIFF
--- a/Dockerfile-bats
+++ b/Dockerfile-bats
@@ -17,11 +17,13 @@ WORKDIR /src/linode-cli
 
 COPY . .
 
-# Preserve the original cli.py file
 RUN mv linodecli/cli.py linodecli/original-cli.py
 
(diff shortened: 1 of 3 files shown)


patthiel:
@Dorthu added a few , thanks!

PR #76, opened 2018-12-05
Cluster create plugin

asauber:
See the updates to README.rst :)

https://github.com/linode/linode-cli/blob/571ceea32159f0167f20e5c2af744b23606fcb21/README.rst#kubernetes-deployment-plugin


linodecli/plugins/k8s-alpha.py
```diff
@@ -0,0 +1,286 @@
...
+        'delete': delete
+    }
+    
+    if parsed.command is None or (parsed.command is None and parsed.help):
+        parser.print_help()
+        print_available_commands(commands)
+        sys.exit(0)
+
+    if parsed.command in commands.keys():
+        commands[parsed.command](args, context)
```
Dorthu: Personally, I'd rewrite this as 

```python
call_command = commands.get(parsed.command)

if call_command is not None:
    call_command(args, context)
```

linodecli/plugins/k8s-alpha.py
```diff
@@ -0,0 +1,286 @@
...
+        nullf = open(os.devnull, 'w')
+        spcall(['terraform'], stdout=nullf)
+    except:
+        print('To create a cluster you must first install Terraform\n'
+              'This command will automatically download and install the Linode provider '
+              'for Terraform.\n'
+              '\nOn macOS with Homebrew: brew install terraform\n'
+              'For other platforms, use your package manager and/or refer to this documentation\n'
+              'https://learn.hashicorp.com/terraform/getting-started/install.html')
+        sys.exit(1)
```
Dorthu: I'd move this check to the top of `call` - as a user, I'd rather know I can't run this command without additional software _before_ I get into writing a specific set of `create` args.

linodecli/plugins/k8s-alpha.py
```diff
@@ -0,0 +1,286 @@
...
+                        help='The path to your private key file which will be used to access Nodes')
+    parser.add_argument('--ssh-public-key', metavar="KEYPATH", type=str, required=False,
+                        default=tf_var_map['ssh_public_key']['default'],
+                        help='The path to your public key file which will be used to access Nodes')
+    parsed, _ = parser.parse_known_args(args)
+
+    # Check if Terraform is installed
+    try:
+        nullf = open(os.devnull, 'w')
+        spcall(['terraform'], stdout=nullf)
```
rmcintosh: you can import `DEVNULL` from `subprocess` for `stdout` instead of doing this

linodecli/plugins/k8s-alpha.py
```diff
@@ -0,0 +1,286 @@
...
+    appdir = safe_make_appdir("k8s-alpha")
+    # Make the terraform project directory if it doesn't exist
+    terrapath = os.path.join(appdir, parsed.name)
+    safe_mkdir(terrapath)
+
+    # Move to the directory
+    os.chdir(terrapath)
+
+    # Generate the terraform file
+    terrafile = open('cluster.tf', 'w')
```
rmcintosh: all of the file writing in this should context manage using `with`

linodecli/plugins/k8s-alpha.py
```diff
@@ -0,0 +1,286 @@
...
+    terrafile.close()
+
+    # Generate terraform args
+    terraform_args = gen_terraform_args(parsed)
+
+    # Run the Terraform commands
+    spcall(['terraform', 'workspace', 'new', parsed.name])
+    call_or_exit(['terraform', 'init'])
+    # TODO: Before running the apply delete any existing Linodes that would
+    # cause the apply to fail.
```
Dorthu: I'm not familiar with Terraform - in what case would existing Linodes cause this apply to fail?

linodecli/plugins/k8s-alpha.py
Dorthu: It appears that this also requires `kubectl` to be installed locally - you should check for that here too

linodecli/plugins/k8s-alpha.py
```diff
@@ -0,0 +1,286 @@
...
+
+    # TODO: Also delete all NodeBalancers and Volumes using the cluster prefix
+
+def quoted_string_or_bare_int(val):
+    if type(val) is int:
+        return val
+    elif type(val) is str:
+        return '"{}"'.format(val)
+    else:
+        return None
```
Dorthu: it looks like this function is only used in string formatting below.  Returning `None` will put the string "None" (without quotes) into the resulting string - you probably want to `return ''` here.

README.rst
```diff
@@ -95,6 +95,67 @@ View your user::
...
+                           The path to your private key file which will be used
+                           to access Nodes
+     --ssh-public-key KEYPATH
+                           The path to your public key file which will be used to
+                           access Nodes
+
+Here's an example use of the command, creating a cluster with 6 2GB Linodes as
+the Nodes.
+
+   linode-cli k8s-alpha create mycluster77 --node-type g6-standard-1 --nodes 6 --master-type g6-standard-4 --region us-east --ssh-private-key $HOME/.ssh/id_rsa --ssh-public-key $HOME/.ssh/id_rsa.pub
```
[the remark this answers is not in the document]
displague: This doesn't send your key - it just instructs Terraform on which key to use when provisioning the nodes.  Terraform will deploy the instances with the public key specified, but it wants to know which private key it will need to use when it uses ssh to reach the nodes.

linodecli/plugins/k8s-alpha.py
```diff
@@ -0,0 +1,286 @@
...
+
+    for varname in tf_var_map.keys():
+        tf_file_parts.append("""variable "{tf_varname}" {{
+  default = {default}
+}}
+""".format(tf_varname=tf_var_map[varname]['name'],
+           default=quoted_string_or_bare_int(tf_var_map[varname]['default'])))
+    
+    tf_file_parts.append("""module "k8s" {{
+  source  = "git@github.com:linode/terraform-linode-k8s.git?ref=for-cli"
```
Dorthu: this looks exceedingly hacky.  Can you promise this will continue working forever?  Is there a more forward-thinking way to do this?

linodecli/plugins/k8s-alpha.py
```diff
@@ -0,0 +1,286 @@
...
+
+  linode_token = "{token}"
```
Dorthu: this looks to be unused

linodecli/plugins/k8s-alpha.py
```diff
@@ -0,0 +1,286 @@
...
+  
+  linode_group = "ka{hashname}-{cluster_name}"
```
Dorthu: this also appears to be unused

README.rst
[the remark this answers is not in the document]
displague: I personally have avoided using the personal key, deferring to the ssh agent support.  But I could see wanting to use different key, not contained in your agent, for this purpose.

Maybe you are running this headless (without an agent), or want to use a unique SSH key pair per cluster.

linodecli/plugins/k8s-alpha.py
```diff
@@ -0,0 +1,286 @@
...
+def gen_terraform_args(parsed):
+    args = []
+    for varname in tf_var_map.keys():
+        args = args + ['-var', "{}={}".format(tf_var_map[varname]['name'], getattr(parsed, varname))]
+    return args
+
+def call_or_exit(*args, **kwargs):
+    ret = spcall(*args, **kwargs)
+    if ret != 0:
+        sys.exit(ret)
```
Dorthu: I assume the called process would've printed an error in this case - is it worth printing something else anyway?

linodecli/plugins/k8s-alpha.py
```diff
@@ -0,0 +1,286 @@
...
+    If we leave the user as kubernetes-admin. Then the configs don't flatten properly.
+    All of them try to create creds for kubernetes-admin.
+    """
+    kubeconfig_fp = os.path.join(terrapath, "{}.conf".format(cluster_name))
+    kubeconfig = open(kubeconfig_fp).read()
+    kubeconfig = kubeconfig.replace('kubernetes-admin', cluster_name)
+
+    kubeconfig_new_fp = os.path.join(terrapath, "{}_new.conf".format(cluster_name))
+    kubeconfig_new = open(kubeconfig_new_fp, 'w')
+    kubeconfig_new.write(kubeconfig)
```
Dorthu: you didn't close either of the file handles you opened here; this is safer to do in a with statement:

```python
kubeconfig_fp = os.path.join(terrapath, "{}.conf".format(cluster_name))
with open(kubeconfig_fp) as f:
    kubeconfig = f.read()

kubeconfig = kubeconfig.replace('kubernetes-admin', cluster_name)

kubeconfig_new_fp = os.path.join(terrapath, "{}_new.conf".format(cluster_name))
with open(kubeconfig_new_fp, 'w') as f:
    f.write(kubeconfig)
```

README.rst
```diff
@@ -95,6 +95,67 @@ View your user::
...
+                           [--ssh-private-key KEYPATH] [--ssh-public-key KEYPATH]
+                           NAME
+
+   positional arguments:
+     NAME                  A name for the cluster.
+
+   optional arguments:
+     -h, --help            show this help message and exit
+     --node-type TYPE      Linode Type ID for cluster Nodes as retrieved with
+                           `linode-cli linodes types`. (default "g6-standard-2")
```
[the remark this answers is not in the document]
displague: The all-purpose default type (say `g6-nanode-1`, `g6-standard-1`, or `g6-standard-20`) may not be a good fit for this use-case. ([Linode APIv4 ID / details gist](https://gist.github.com/displague/3c1728fdf4ff2bacf6960a9b8c6ea73f))

linodecli/plugins/k8s-alpha.py
```diff
@@ -0,0 +1,286 @@
...
+
+    return kubeconfig_new_fp
+
+def safe_make_appdir(appname):
+    if sys.platform == 'win32':
+        appdir = os.path.join(os.environ['APPDATA'], appname)
```
Dorthu: I'm not very familiar with Windows these days, but the internet suggests this will return `AppData\Roaming`.  Either way, `os.environ.get('APPDATA')` won't raise if it isn't set (although I'm not sure what good behavior would be if it wasn't set)

README.rst
```diff
@@ -95,6 +95,67 @@ View your user::
...
+     --nodes COUNT         The number of Linodes to deploy as Nodes in the
+                           cluster. (default 3)
+     --master-type TYPE    Linode Type ID for cluster Master Nodes as retrieved
+                           with `linode-cli linodes types`. (default
+                           "g6-standard-2")
+     --region REGION       The Linode Region ID in which to deploy the cluster as
+                           retrieved with `linode-cli regions list`. (default
+                           "us-west")
```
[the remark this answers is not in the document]
displague: It would be good to pick up the already chosen default region.

README.rst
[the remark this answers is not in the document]
jfrederickson: It may be worth mentioning this in the documentation, FWIW; this is the second time I've seen this brought up.

README.rst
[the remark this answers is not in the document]
displague: If https://github.com/linode/linode_api4-python/issues/147 were to use TOML, the `linode-cli` settings could have a sub-section per `linode-cli` plugin (for layers of overriding defaults).

linodecli/plugins/k8s-alpha.py
```diff
@@ -0,0 +1,286 @@
...
+    A cluster hashname is the first 9 characters of a SHA256 digest encoded in base64
+
+    This is used as a compact way to uniquely identify a cluster's Linode resources.
+    It's also stateless! If a user loses their terraform file and wishes to
+    delete a cluster they can still do so.
+    """
+    hashname = base64.b64encode(hashlib.sha256(name.encode('utf8')).digest())
+    hashname = hashname.decode('utf8')
+    hashname = hashname.replace(r'+', r'')
+    hashname = hashname.replace(r'/', r'')
```
Dorthu: These work without being regex strings (i.e. `replace('+', '')`)

linodecli/plugins/k8s-alpha.py
rmcintosh: https://github.com/linode/linode-cli/pull/76#discussion_r239800753

README.rst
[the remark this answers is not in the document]
Dorthu: 🙃 TIL

linodecli/plugins/k8s-alpha.py
Dorthu: fair

README.rst
```diff
@@ -95,6 +95,67 @@ View your user::
...
 
+Kubernetes Deployment Plugin
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A plugin is included that allows you to deploy a Kubernetes cluster on Linode!
+
+This Kubernetes cluster is integrated with Linode in a number of ways:
+
+   * When you deploy a LoadBalancer-type service a Linode NodeBalancer will be
+     automatically created and managed for the Pods backing that service.
```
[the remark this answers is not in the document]
asauber: will do

linodecli/plugins/k8s-alpha.py
asauber: In subprocess `DEVNULL = -3`, in other operating environments, the "null file descriptor" may not be -3. This is more platform independent.

linodecli/plugins/k8s-alpha.py
asauber: I agree

README.rst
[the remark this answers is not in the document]
asauber: Region I will definitely pull from the cli config (had a fleeting thought to do this last week and forgot to do it)

README.rst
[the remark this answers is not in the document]
asauber: For both Node and Master I don't think using the user's default Linode type is a good choice.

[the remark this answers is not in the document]
asauber: A better thing to do here would be to query for Linode types and look for the first one with at least 2 cores and 4GB

README.rst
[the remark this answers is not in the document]
asauber: I'm going to do some testing and see how smart Terraform's agent is about inferring the location of a privkey from a public key path. Maybe this isn't needed. If it is, it will come with an explanation.

README.rst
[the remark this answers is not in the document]
asauber: Will do

linodecli/plugins/k8s-alpha.py
asauber: And exit otherwise. Safe access is always better I agree.

linodecli/plugins/k8s-alpha.py
asauber: Agree on both counts.

linodecli/plugins/k8s-alpha.py
asauber: In some cases I've seen Terraform fail to provision a Linode and not have it stored in state (to be re-provisioned on a subsequent run). In this case the user is stuck until the manually delete the Linode. It's rare.

linodecli/plugins/k8s-alpha.py
```diff
@@ -0,0 +1,286 @@
...
+                        default=tf_var_map['region']['default'],
+                        help='The Linode Region ID in which to deploy the cluster as retrieved with '
+                             '`linode-cli regions list`. (default "us-west")')
+    parser.add_argument('--ssh-private-key', metavar="KEYPATH", type=str, required=False,
+                        default=tf_var_map['ssh_private_key']['default'],
+                        help='The path to your private key file which will be used to access Nodes')
+    parser.add_argument('--ssh-public-key', metavar="KEYPATH", type=str, required=False,
+                        default=tf_var_map['ssh_public_key']['default'],
+                        help='The path to your public key file which will be used to access Nodes')
+    parsed, _ = parser.parse_known_args(args)
```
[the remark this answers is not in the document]
asauber: I used `parse_known_args` because I originally passed all remaining args to `terraform apply`. This can be removed.

linodecli/plugins/k8s-alpha.py
```diff
@@ -0,0 +1,286 @@
...
+
+    # We're done! We have merged the user's kubeconfigs.
+    # So, the user should be able to run something like
+    # `kubectl get pods --all-namespaces`
+    # and see the Linode CSI, CCM, and ExternalDNS controllers
+
+def delete(args, context):
+    parser = argparse.ArgumentParser("{} create".format(plugin_name), add_help=True)
+    parser.add_argument('name', metavar='NAME', type=str,
+                        help="A name for the cluster.")
```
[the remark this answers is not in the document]
asauber: Yes. This `delete` command should also check for the presence of `terraform`, which I'll factor out.

linodecli/plugins/k8s-alpha.py
asauber: Great point

linodecli/plugins/k8s-alpha.py
Dorthu: I'll see myself out then 🤦‍♂️

linodecli/plugins/k8s-alpha.py
Dorthu: I mean the `linode_group` variable defined on this line doesn't appear to be referenced elsewhere in this file

asauber:
Link to CSI and CCM repo in docs.

asauber:
Output some helpful information after the cluster is deployed.

linodecli/plugins/k8s-alpha.py
rmcintosh: other operating environments are stupid

rmcintosh: can we context manage this one too, in that case?

linodecli/plugins/k8s-alpha.py
asauber: You're right that these don't need to be raw strings

linodecli/plugins/k8s-alpha.py
asauber: Right, because it's a variable for Terraform. This is a string/text file

linodecli/plugins/k8s-alpha.py
asauber: It already has this semantics, so I'm going to leave it.

linodecli/plugins/k8s-alpha.py
Dorthu: this is part of a multi-line string 😓 missed that, sorry

PR #79, opened 2018-12-06
Don't define blank values in config when the user declines optional defaults

rmcintosh:
Currently, if you choose to ignore setting defaults during `configure`, you end up with a config that looks like this:

```
[USERNAME]
    default1 =
    default2 =
    default3 =
```

This breaks e.g. creating an empty Linode, since we try to pass e.g. the Image to the API, but it's a blank string (instead of not passing it at all). Removing these blank defaults fixes the issue, but simply not defining blank defaults in the config, when a user declines an optional default, fixes the application bug.

### DIFF
--- a/linodecli/configuration.py
+++ b/linodecli/configuration.py
@@ -147,9 +147,10 @@ on your account to work correctly.""".format(TOKEN_GENERATION_URL))
             self.config.add_section(username)
 
         for k, v in config.items():
-            self.config.set(username, k, v)
-            if is_default:
-                self.config.set('DEFAULT', k, v)
+            if v:
+                self.config.set(username, k, v)
+                if is_default:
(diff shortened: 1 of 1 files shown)


linodecli/configuration.py
```diff
@@ -147,9 +147,10 @@ def configure(self, username=None):
             self.config.add_section(username)
 
         for k, v in config.items():
-            self.config.set(username, k, v)
-            if is_default:
-                self.config.set('DEFAULT', k, v)
+            if v:
```
rmcintosh: alternately: `if len(v)` if we are worried about future boolean options being possibly misinterpreted

rmcintosh:
An alternate version of this would have the blank config values remain, but have the CLI not use them if they're blank - but I kind of think that sounds dumb

PR #83, opened 2018-12-08
[Docs] Example of custom output via mustache

ilsken:
Add an example of custom output via mustache command. Not crucial or anything but I figure it's nice to throw in there while we're looking at output formatters 🤷‍♂️


Dorthu:
This is a cool idea, but I'd feel better about it if there was an example that actually usable (i.e. a provided template to create some useful output for a common command).

asauber:
I agree, there are a number of tools that consume json on stdin. An example template would make this more useful.

Dorthu:
Closing because this hasn't been updated in a long time.  Please re-open if you're still interested.

PR #85, opened 2018-12-10
add: (k8s-alpha) Warn about non-running ssh-agent

asauber:
Check if the provided pubkey has a corresponding private key loaded in a running ssh-agent.


asauber:
Tested with both Python 2 and 3 on both macOS and Ubuntu 18.04

PR #87, opened 2018-12-10
hint at alpha-k8s dependencies prior to command execution

sarahelizgray:
@Dorthu and @asauber it might be nice to talk about dependencies before command execution, as per Howie.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -105,7 +105,7 @@ View your user::
 Kubernetes Deployment Plugin
 """"""""""""""""""""""""""""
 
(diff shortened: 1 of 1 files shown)


asauber:
suggest "this command" as it is one command with subcommands

sarahelizgray:
@asauber I went with "these" b/c I assume deletion and other commands within the alpha-k8s plugin have the same dependencies. Does that sway you? I can go either way.

sarahelizgray:
@asauber and @Dorthu, can I get a merge on this?

PR #90, opened 2018-12-26
fix: Allow merging of settings when token is not present in config

alxbl:
It turns out that using the environment variable to store the token can fail because the check for merging the configured settings and the command line supplied settings relies on `token` being present in the configuration. I hadn't noticed because I had already done a `configure` beforehand which stored a valid token in the `DEFAULT` user.

This PR fixes that issue by letting the configuration merge happen when the environment variable is defined.

Sorry for not catching this sooner! If you'd like me to open an issue for tracking purposes, let me know.

Cheers,

### DIFF
--- a/linodecli/configuration.py
+++ b/linodecli/configuration.py
@@ -63,7 +63,7 @@ class CLIConfig:
         if not username:
             username = "DEFAULT"
 
-        if not self.config.has_option(username, 'token') and not os.environ.get(ENV_TOKEN_NAME, None:
+        if not self.config.has_option(username, 'token') and not os.environ.get(ENV_TOKEN_NAME, None):
             print("User {} is not configured.".format(username))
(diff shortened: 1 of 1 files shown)


linodecli/configuration.py
```diff
@@ -63,7 +63,7 @@ def update(self, namespace, username=None):
         if not username:
             username = "DEFAULT"
 
-        if not self.config.has_option(username, 'token'):
+        if not self.config.has_option(username, 'token') and not os.environ.get(ENV_TOKEN_NAME, None:
```
Dorthu: pretty sure you're missing a `)` here

alxbl: Nice catch. I must've accidentally removed it right before committing :S I'll update the PR.

Sorry about that.

Issue #77, opened 2018-12-06
Use $XDG_CONFIG_HOME for config files


tarqd:
I think it'd be better if `linode-cli` respected the `$XDG_CONFIG_HOME ` variable and defaulted to `$HOME/.config/linode-cli` as the config dir if it's not set.

Personally, I don't like my home directory getting cluttered and this variable seems to be the standard for newer software unix-y systems. 

Related package: https://pypi.org/project/xdg/

It's fairly small, just checks the env variables and has sane defaults if the variables are not set.

Reference: https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

lgarber-akamai:
Closing since the relevant PR has already been merged.

PR #94, opened 2019-01-15
Xdg environment

patthiel:
* Implements https://github.com/linode/linode-cli/issues/77
* The new path to the config file is the `XDG_CONFIG_HOME`, but it will fallback to the legacy path if that config file is not present. This will makes this a non-breaking change.

#### Testing 

1. Re-install the CLI from this branch
2. Preserve your original cli config (`~/.linode-cli`) somewhere
3. Move `~/.linode-cli` out of your home folder see that your new config is saved to `XDG_CONFIG_HOME` (defaults to `~/.config/linode-cli`). 
3. Move your original cli config back to the home folder, remove the `~/.config/linode-cli` config.
4. Run the CLI it should use your legacy config path


requirements.txt
```diff
@@ -3,3 +3,4 @@ colorclass
 requests
 PyYAML
 enum34
+xdg
```
Dorthu: This module looks simple enough, but is it really worth adding another external dependency for what is essentially a one-liner?

ilsken: Personally I'd opt to use the module, if they add more OS' etc we get it for free and the dep is really small in the scheme of things (all the dub dependencies of requests, pyaml, etc)

patthiel: I don't feel strongly about this. It is a super small module and adding it could result in less maintenance and better defaults across different OS's over time. @Dorthu deferring to you on this one.

Dorthu: Yeah, you've convinced me, this is fine

Dorthu:
Let me test this locally.  This is one of two CLI PRs that is on my shortlist, and I promise I'll get to soon.

patthiel:
@Dorthu that's fair. I've pushed a commit that makes the new default location for cli configs to be in `~/.config/linode-cli` (which is the default XDG_ENVIRONMENT). It will, same as before, fall back to the ~/.linode-cli` path if it's not found.

linodecli/configuration.py
```diff
@@ -17,8 +18,10 @@
 
 ENV_TOKEN_NAME='LINODE_CLI_TOKEN'
 
-CONFIG_DIR = os.path.expanduser('~')
-CONFIG_NAME = '.linode-cli'
+LEGACY_CONFIG_DIR = os.path.expanduser('~')
+LEGACY_CONFIG_NAME = '.linode-cli'
+CONFIG_DIR = os.environ.get('XDG_CONFIG_HOME', "{}/{}/".format(os.path.expanduser('~'), '.config'))
```
Dorthu: I believe this trailing `/` is redundant

Dorthu:
Is there still interest in seeing this merged?

patthiel:
@Dorthu I updated this PR to create `~/.config` if that does not exist already. Also removed the extraneous `/` you mentioned.. I think that addresses everything. Old config files are still used if they exist, otherwise the new XDG config path is where configs will be written (defaulting to `~/.config` if env var is not present).

PR #95, opened 2019-01-15
Fix spec fetching on alternative environments

patthiel:
* Removes fetching the open api spec from inside the container.
* Instead, save the openapi spec outside the container context and pass in the filename as a build argument.
* Feel free to ping me for further instructions on testing this if needed

### DIFF
--- a/test/nodebalancers/nodebalancers.bats
+++ b/test/nodebalancers/nodebalancers.bats
@@ -86,13 +86,19 @@ export nodebalancerCreated="[0-9]+,balancer[0-9]+,us-east,nb-[0-9]+-[0-9]+-[0-9]
 }
 
 @test "it should add a node to the configuration profile" {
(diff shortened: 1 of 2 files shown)


test/nodebalancers/nodebalancers.bats
```diff
@@ -86,13 +86,19 @@ export nodebalancerCreated="[0-9]+,balancer[0-9]+,us-east,nb-[0-9]+-[0-9]+-[0-9]
...
+	nodeImage=$1
+
 	nodeIp=$(linode-cli linodes create \
 	     --root_pass aComplex@Password \
 	     --booted true \
 	     --region us-east \
 	     --type g6-standard-2 \
 	     --private_ip true \
-	     --image linode/arch \
+	     --image $nodeImage \
```
Dorthu: This bit looks unrelated.  Was this test broken?

PR #99, opened 2019-02-13
Add Power Status tests, Fix broken tags test

patthiel:
* Does the thing in the title, also adds a test to linodes to create a image-less Linode and assert that it does not boot.
* Updates the tags assertion due to the increase in max length for tags

### DIFF
--- a/test/linodes/linodes.bats
+++ b/test/linodes/linodes.bats
@@ -49,6 +49,7 @@ load '../common'
         --format 'id,region,type'
 
     assert_output --regexp "[0-9]+,us-east,g6-standard-2"
+    run removeLinodes
 }
 
 @test "it should fail to create a linode without a root_pass" {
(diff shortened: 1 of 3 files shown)


test/linodes/linodes.bats
```diff
@@ -65,13 +66,33 @@ load '../common'
...
+        --no-defaults \
+        --label='cli-2' \
+        --type=$linode_type \
+        --region=$linode_region \
+        --root_pass $random_pass
+    local linode_id=$(linode-cli linodes list --format="id" --text --no-headers)
+
+    until [ $(linode-cli linodes view $linode_id --format="status" --text --no-headers) = "offline" ]; do
+        echo 'still setting up'
+    done
```
Dorthu: This loop should probably have a timeout.  Also, adding a sleep per iteration would be friendly 😄

test/linodes/linodes.bats
```diff
@@ -65,13 +66,33 @@ load '../common'
...
+    local linode_region=$(linode-cli regions list --format="id"  --text --no-headers | xargs | awk '{ print $1 }')
+    run linode-cli linodes create \
+        --no-defaults \
+        --label='cli-2' \
+        --type=$linode_type \
+        --region=$linode_region \
+        --root_pass $random_pass
+    local linode_id=$(linode-cli linodes list --format="id" --text --no-headers)
+
+    until [ $(linode-cli linodes view $linode_id --format="status" --text --no-headers) = "offline" ]; do
```
Dorthu: This test would mean more if you check that the status wasn't still "provisioning"

test/linodes/power-status.bats
```diff
@@ -0,0 +1,102 @@
...
+
+    assert_success
+    assert_output "provisioning"
+}
+
+@test "it should create a linode and boot" {
+    run createLinode
+    local linode_id=$(linode-cli linodes list --format="id" --text --no-headers)
+
+    until [ $(linode-cli linodes view $linode_id --format="status" --text --no-headers) = "running" ]; do
```
Dorthu: Again, I'd prefer a `!= "provisioning"` here

test/linodes/power-status.bats
```diff
@@ -0,0 +1,102 @@
...
+        echo 'still provisioning'
+    done
```
Dorthu: And a sleep in the loop would be a cool addition too

patthiel:
fair feedback. I added timeouts of 180 seconds for the until loops and added sleeps of 5 seconds between each request to be nicer, let me know if i should be more generous with that time between requests. 

As far as the method in the tests. I'm not really a fan of the "wait for this thing to _not_ be in state T,U,V then make an assertion that it's in state Z". That pattern could result in more flakey tests, because there's no guarantee that they'll be in state Z after the initial "NOT in T,U,V" conditionals. They could, eventually be in state Z and allowing for some temporary intermediate state that we don't care about that eventually transitions to the end state is ok, so long as we have a *timeout*, which I've now added (thx to you). What if more steps between T,U,V are added later. They'll break these tests, even though all i care about is that it gets to step Z

test/linodes/linodes.bats
Dorthu: This loop still has no timeout or rate limit

patthiel: good catch, added!

patthiel:
Think i addressed all the feedback here. Thanks for reviewing!

PR #101, opened 2019-02-15
Add Rebuild, Resize, Backup Tests

patthiel:
* Adds tests for Rebuilding, Resizing, and creating backups.
* Also fixed a breaking stackscripts test
* Also added a helper to create a linode and wait for running

## To Test:

```
brew install bats-core
bats test/linodes/rebuild.bats test/linodes/resize.bats test/linodes/backups.bats
```

### DIFF
--- a/test/common.bash
+++ b/test/common.bash
@@ -86,13 +86,11 @@ createLinodeAndWait() {
     SECONDS=0
     until [ $(linode-cli linodes view $linode_id --format="status" --text --no-headers) = "running" ]; do
         echo 'still provisioning'
-        sleep 3 # Wait 3 seconds between requests
+        sleep 5 # Wait 5 seconds before checking status again, to rate-limit ourselves
         if [[ "$SECONDS" -eq 180 ]];
         then
             assert_failure # Fail test, linode did not boot in time
             break
         fi
-        # Wait 3 seconds before checking status again, to rate-limit ourselves
-        sleep 3
     done
 }
--- a/test/common.bash
+++ b/test/common.bash
@@ -83,10 +83,15 @@ createLinodeAndWait() {
 
     local linode_id=$(linode-cli linodes list --format id --text --no-header | head -n 1)
 
-
+    SECONDS=0
     until [ $(linode-cli linodes view $linode_id --format="status" --text --no-headers) = "running" ]; do
         echo 'still provisioning'
-
+        sleep 3 # Wait 3 seconds between requests
+        if [[ "$SECONDS" -eq 180 ]];
+        then
+            assert_failure # Fail test, linode did not boot in time
+            break
+        fi
         # Wait 3 seconds before checking status again, to rate-limit ourselves
         sleep 3
     done
--- a/test/common.bash
+++ b/test/common.bash
@@ -74,3 +74,20 @@ removeAll() {
 removeUniqueTag() {
     run bash -c "linode-cli tags delete $uniqueTag"
 }
+
+createLinodeAndWait() {
+    local default_plan=$(linode-cli linodes types --text --no-headers --format="id" | xargs | awk '{ print $1 }')
+    local linode_type=${1:-$default_plan}
+    run bash -c "linode-cli linodes create --type=$linode_type --region us-east --image=$test_image --root_pass=$random_pass"
+    assert_success
+
+    local linode_id=$(linode-cli linodes list --format id --text --no-header | head -n 1)
+
+
+    until [ $(linode-cli linodes view $linode_id --format="status" --text --no-headers) = "running" ]; do
+        echo 'still provisioning'
+
+        # Wait 3 seconds before checking status again, to rate-limit ourselves
+        sleep 3
+    done
+}
--- a/test/linodes/backups.bats
+++ b/test/linodes/backups.bats
@@ -11,12 +11,13 @@ setup() {
     export timestamp=$(date +%s)
     clean_linodes="false"
     linode_id=$(linode-cli linodes list --format id --text --no-header | head -n 1)
+    snapshot_label="test_snapshot1"
 }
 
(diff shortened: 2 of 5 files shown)


test/common.bash
```diff
@@ -74,3 +74,25 @@ removeAll() {
...
+    run bash -c "linode-cli linodes create --type=$linode_type --region us-east --image=$test_image --root_pass=$random_pass"
+    assert_success
+
+    local linode_id=$(linode-cli linodes list --format id --text --no-header | head -n 1)
+
+    SECONDS=0
+    until [ $(linode-cli linodes view $linode_id --format="status" --text --no-headers) = "running" ]; do
+        echo 'still provisioning'
+        sleep 3 # Wait 3 seconds between requests
+        if [[ "$SECONDS" -eq 180 ]];
```
Dorthu: this might be me being naive, but is anything incrementing SECONDS here?

test/common.bash
```diff
@@ -74,3 +74,25 @@ removeAll() {
...
+        then
+            assert_failure # Fail test, linode did not boot in time
+            break
+        fi
+        # Wait 3 seconds before checking status again, to rate-limit ourselves
+        sleep 3
```
Dorthu: This is a double-sleep - is that what you wanted?

test/linodes/backups.bats
```diff
@@ -0,0 +1,105 @@
...
+    if [ "$clean_linodes" = "true" ];
+    then
+        run removeLinodes
+    fi
+}
+
+@test "it should create a linode with backups disabled" {
+    run createLinode
+    local linode_id=$(linode-cli linodes list --format id --text --no-header | head -n 1)
+    run linode-cli linodes backups-cancel $linode_id --text
```
Dorthu: This is a kinda lousy way to assert that backups are enabled.  The CLI supports showing this. .but the attribute name is lousy: `linode-cli linodes list --format 'id,enabled'` shows you all your Linodes and if their backups are enabled.

test/linodes/backups.bats
```diff
@@ -0,0 +1,105 @@
...
+@test "it should enable backups" {
+    run linode-cli linodes backups-enable $linode_id \
+        --text \
+        --no-headers
+    assert_success
+
+    run linode-cli linodes backups-cancel $linode_id \
+        --text \
+        --no-headers
+    assert_success
```
Dorthu: Again, you could just check if backups are enabled

test/linodes/rebuild.bats
```diff
@@ -0,0 +1,90 @@
...
+        --root_pass=$random_pass \
+        --text \
+        --no-headers \
+        $linode_id
+
+    assert_success
+
+    # Wait until rebuilding
+    SECONDS=0
+    until [ $(linode-cli linodes view $linode_id --format="status" --text --no-headers) = "rebuilding" ]; do
```
Dorthu: fwiw, this should be immediate

test/linodes/rebuild.bats
```diff
@@ -0,0 +1,90 @@
...
+        fi
+    done
+
+    run linode-cli linodes view $linode_id \
+    	--format="image" \
+    	--text \
+    	--no-headers
+
+    assert_success
+    assert_output "$rebuild_image"
```
Dorthu: I'm not objecting, but the bottom half of this test only works because the Linode was initially running, and that initial state should be documented to make it clear that the test depends on it.

test/linodes/resize.bats
```diff
@@ -0,0 +1,109 @@
...
+		--no-headers \
+		$linode_id
+
+	assert_success
+
+	# Wait for status = "Resizing"
+    SECONDS=0
+	until [ $(linode-cli linodes view $linode_id --format="status" --text --no-headers) = "resizing" ]; do
+        echo 'waiting for resize to start'
+        sleep 5
```
Dorthu: this is a `sleep 5` whereas elsewhere in this PR you use `sleep 3` - any reason in particular?

test/linodes/resize.bats
```diff
@@ -0,0 +1,109 @@
...
+
+	# Wait for offline status.
+	# Linodes that are resized do not boot automatically
+    SECONDS=0
+    until [ $(linode-cli linodes view $linode_id --format="status" --text --no-headers) = "offline" ]; do
+        echo 'still resizing'
+
+		# Check for resizing completion every 15 seconds
+        sleep 15
+        if [[ "$SECONDS" -eq 600 ]];
```
Dorthu: The test suite is going to be a lot slower with these additions

test/common.bash
patthiel: Type `echo $SECONDS` in your shell, then type `SECONDS=0` then `echo $SECONDS` again 😉

test/common.bash
patthiel: nope! removing and going to make `sleep ` times a constant.

test/linodes/backups.bats
patthiel: awesome, i was thinking the same when i wrote it, but at the time didn't know about the above. updating.

test/linodes/backups.bats
patthiel: 👍

test/linodes/resize.bats
patthiel: @Dorthu indeed. In the networking tests PR, i neglected to mention that I did add a conditional check in the long running tests to see if the env variable `CI_BUILD=TRUE`.. When set to false it will skip the long-running tests. I could make that more explicit like, `SKIP_LONG_TESTS=TRUE` and add it as a jenkins parameterized option.

test/common.bash
Dorthu: TIL

PR #105, opened 2019-02-26
Do not print warning when API Spec's patch version is greater than CLI Spec

patthiel:
* This PR changes the version comparison warning to only print the warning when the spec version returned by the API is a major or minor version greater than the CLI spec.

Example:
```
API Spec - 4.0.5 
CLI Spec - 4.0.6  # CLI was printing a warning prior to this PR, now it will not complain
```

### DIFF
--- a/linodecli/cli.py
+++ b/linodecli/cli.py
@@ -3,7 +3,7 @@ Responsible for managing spec and routing commands to operations.
 """
 from __future__ import print_function
 
-from distutils.version import LooseVersion
+from distutils.version import StrictVersion, LooseVersion
 import json
 import os
 import pickle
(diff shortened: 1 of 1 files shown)


patthiel:
@Dorthu I confirmed that `make build && make install` work just fine with this change.

patthiel:
What command did you run in order to hit that? I can't reproduce locally, and the dockerfile builds the CLI successfully:
`docker build -f Dockerfile-bats -t cli .`
`docker run cli linode-cli` outputs what i would expect

Dorthu:
I built it with a local copy of the spec, whose version is "DEVELOPMENT".  This is done internally to test that the CLI builds against unreleased specs.

patthiel:
Pushed a change to handle the above.

patthiel:
Addressed some more of your concerns about how strict my handling of non-standard version values was. 

I have it now parsing the versions provided using StrictVersion. If they throw a ValueError the except will handle falling back to using the original values and pass them into `LooseVersion`.

PR #107, opened 2019-03-06
Allow plugins to provide arbitrary filter headers

Dorthu:
This is for @jfrederickson

Plugins sometimes need to use more complex filtering than the CLI
provides by default.  This allows plugins only to specify filter headers
as Python dicts, and overrides whatever was passed in the CLI command.


patthiel:
LGTM

Issue #92, opened 2019-01-03
k8s-alpha create assumes that public ssh key contains filename


kengu:
**Note**: This error is only reproduced on mac.

The k8s-alpha plugin assumes that the public ssh key contains the filename, which I guess is not stable across different OS implementations. On mac, the default comment `user@hostname`, which does not match the key filename which the script checks for. As a result, the create command generates the warning "Your ssh private key must be added to your ssh-agent." even if you have added the key. The workaround I used was to generate a key that the script accepted was

```
> ssh-keygen -b 4096 -t rsa -C "id_rsa"`
> ssh-add -K ~/.ssh/id_rsa
> linode-cli k8s-alpha create k8s --ssh-public-key ~/.ssh/id_rsa.pub
```

See [Generating a new SSH key and adding it to the ssh-agent](https://help.github.com/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent/#platform-mac)

asauber:
Hi @kengu, thanks for the report

The plugin doesn't assume that the key data contain the filename, but it does assume that `ssh-add -l` output filepaths.

For example here's that output on my system:
```
asauber@linaret2 linode-cloud-controller-manager # ssh-add -l
2048 SHA256:Qcl/i4Yd9E7vXQbiJ98rFTHssJ1TVGKyaRYsjstWypg /Users/asauber/.ssh/clidemo (RSA)
4096 SHA256:Pv0U9hFqENpF5cW0dqrsr5zA7XYgHaNtYPRpzEMe0Hg /Users/asauber/.ssh/k8sdemo (RSA)
2048 SHA256:OcziYNj7HWb1yZINdJxXy59dTjDqZ1IwYdul6ddTeVc /Users/asauber/.ssh/id_rsa (RSA)
4096 SHA256:DpwXaeZ25N5D9CaFLy4aa5pumxvOVhNZ5qW3xI6wOyY /Users/asauber/.ssh/testkey (RSA)
```

asauber:
Does yours only output the comment where mine outputs the filepath? Maybe we're running different versions of `ssh-add`. I tested that the "filepath output" behavior was consistent on a few Linux distributions before using it.

There must be some differences between our OS configurations or our version of `ssh-add`

Unfortunately `ssh-add` doesn't have a version flag, so I'm skimming through `strings $(which ssh-add)` at the moment.

kengu:
[...]

PR #108, opened 2019-03-06
Dynamically set default Linode type for deployed K8s masters

jfrederickson:
**NOTE:** Requires https://github.com/linode/linode-cli/pull/107

Currently, we set hardcoded instance types for both nodes and
masters. However, the linode-cli lets you specify a default type for
newly created Linodes.

Ideally, we would like to respect that choice, however K8s masters do
require 2 VCPUs at minimum. Since the available plans are available
through the API, we can look up the user's configured default type to
see if it meets this constraint. If it does, it's fine to use that. If
it doesn't, we pick the smallest plan that meets this constraint. (In
this case the user's default is smaller than the minimum, so they're
probably looking to deploy smaller Linodes if possible.)

Command output of this running after changing my default Linode type to g6-standard-8: https://gist.github.com/jfrederickson/02396ab8a0de57b4bd8370ebc252fc13

It's worth noting that if we ever add functionality to this plugin that modifies existing clusters, and the API returns a different instance type from that `type-view` request, this could result in the master getting destroyed and recreated. I suspect we may never add that functionality to this plugin in particular, but I thought it was worth mentioning.

(Also let me know if you want me to get a bit more structured with that error handling...)

### DIFF
--- a/linodecli/plugins/k8s-alpha.py
+++ b/linodecli/plugins/k8s-alpha.py
@@ -44,7 +44,7 @@ def create_varmap(context):
     tf_var_map = {
         'node_type': {
             'name': 'server_type_node',
-            'default': context.client.config.config.get('DEFAULT', 'type', fallback='g6-standard-2'),
+            'default': requested_plan_with_fallback(context),
         },
         'nodes': {
             'name': 'nodes',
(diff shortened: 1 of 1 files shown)


asauber:
Added a commit to this for Python2 support

Dorthu:
@asauber @jfrederickson when you you want this to ship?

asauber:
We have another PR in the works for #92, so I would say after that is merged or if an earlier release is needed.

PR #109, opened 2019-03-08
Added multi-user support

### LINKED ISSUE #103: Multiple Account
How to manage multiple account using linode-cli?

I can add multiple account using `linode-cli configure`, but there's no way to run linode-cli using specific account

Dorthu:
Closes #103

This is implemented as described in #103 - feedback is appreciated!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -113,11 +113,11 @@ run the *configure* command::
 Suppressing Defaults
 """"""""""""""""""""
 
-If you configured default values for `image`, `region`, and Linode `type`, they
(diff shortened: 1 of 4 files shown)


patthiel:
The default usage / help information makes no mention of the commands to show users, add users , remove users. We should add them in the available commands, maybe in a separate column to the API commands?

patthiel:
This didn't break my CLI in my testing, but my config is pretty vanilla with no defaults selected, just a token. 

tested adding/listing users/running commands as specific users, updating my default user.

PR #110, opened 2019-03-12
coaching language for if you don't use the default ssh key name, closes #92

### LINKED ISSUE #92: k8s-alpha create assumes that public ssh key contains filename

sarahelizgray:


linodecli/plugins/k8s-alpha.py
```diff
@@ -110,8 +110,9 @@ def create(args, context):
                         default=tf_var_map['ssh_public_key']['default'],
                         help='The path to your public key file which will be used to access Nodes '
                              'during initial provisioning only! The keypair _must_ be added to an '
-                             'ssh-agent (default {})'.format(
-                                 tf_var_map['ssh_public_key']['default']))
+                             'ssh-agent. If you don\'t use id_rsa as your private key name, use '
+                             'the flag --ssh-public-key and supply your public key path. If you '
+                             'use id_rsa as your key name, omit the flag. (default {}).'.format(tf_var_map['ssh_public_key']['default']).)
```
Dorthu: There's a trailing period here causing a syntax error

sarahelizgray: fixed

sarahelizgray:
@Dorthu, assuming it meets with @asauber's expectations, yes!

linodecli/plugins/k8s-alpha.py
```diff
@@ -252,10 +253,10 @@ def print_ssh_agent_warning(parsed, with_agent_start=False):
     agent_start = ''
     if with_agent_start:
         agent_start = "eval $(ssh-agent) && "
-    print("Your ssh private key must be added to your ssh-agent.\n"
-          "Please run this command:\n\n{}ssh-add {}".format(
-            agent_start,
-            parsed.ssh_public_key.replace('.pub', '')))
+    print("If you wish to omit specifying the public key path, you must use \n"
```
asauber: This message might be shown even if the user has provided `--ssh-public-key` but that key is not added to their agent. So it's not accurate to assume that they've omitted the public key path.

The instructions dropped `eval $(ssh-agent) && ` and placed commands in the middle of a line, which we avoid elsewhere in this plugin.

It's also not accurate to request an `ssh-add` and then say "or" use the flag --ssh-public-key, a user might want to do both (in fact, they have to if they use --ssh-public-key and don't have an agent running or don't have their key added)

sarahelizgray:
I'm good with @asauber's edits.

PR #117, opened 2019-03-25
Added support for plugin configuration

Dorthu:
Plugins may want to store persistent configuration data on a per-user
bases.  Instead of having them manage that themselves, this allows
plugins to use the Linode CLI's config in a uniform manner to store
persistent data.

All plugin config items' keys begin with `plugin-` followed by the
plugin's name.  Keys are stored per-user in the config, so multiple
users may have different settings configured.

This is not presently in use, but it might come in handy for some
upcoming plugins that may or may not be in the works ;)

### DIFF
--- a/linodecli/configuration.py
+++ b/linodecli/configuration.py
@@ -137,6 +137,30 @@ class CLIConfig:
 
     # plugin methods - these are intended for plugins to utilize to store their
     # own persistent config information
+    def get_value(self, key):
+        """
+        Retrieves and returns an existing config value for the current user.  This
+        is intended for plugins to use instead of having to deal with figuring out
+        who the current user is when accessing their config.
+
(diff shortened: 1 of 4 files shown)


linodecli/configuration.py
```diff
@@ -129,7 +133,94 @@ def set_default_user(self, username):
...
+        :rtype: any
+        """
+        username = self.username or self.default_username()
+
+        if not self.config.has_option(username, key):
+            return None
+
+        return self.config.get(username, key)
+
+    def plugin_set_value(self, plugin_name, key, value):
```
asauber: I think having name as a parameter here leaves too much up to the plugin. Why not have plugins register their name in a global on invoke and then read that here.

What do you think?

PR #121, opened 2019-04-16
Adds `ssh` plugin

Dorthu:
This replaces #72.

Adds a new plugin, `ssh`, which is invoked as follows:

```bash
linode-cli ssh root@my-linode-label
```

The username piece is optional and defaults to the user invoking the CLI
(as normal ssh behavior).  This is basically a proxy for finding a
public IP address to ssh into the Linode with by searching for a Linode
with the given label.

If no Linode with the given label exists but other Linodes are found
that are close, those labels are output as a suggestion.

All arguments after the `[USERNAME@]LINODE_LABEL` are passed on to ssh
as normal.

Big thanks to @displague and @jfrederickson for their feedback regarding
this plugin.

### DIFF
--- a/linodecli/plugins/ssh.py
+++ b/linodecli/plugins/ssh.py
@@ -40,6 +40,8 @@ def call(args, context):
                         help="The label of the Linode to SSH into, optionally with "
                              "a username before it in USERNAME@LABEL format.  If no "
                              "username is given, defaults to the current user.")
+    parser.add_argument('-6', action='store_true',
+                        help="If given, uses the Linode's SLAAC address for SSH.")
 
     parsed, args = parser.parse_known_args(args)
 
@@ -56,8 +58,7 @@ def call(args, context):
             "linodes", "list", filters={'label': {"+contains": label}})
 
     if result != 200:
-        # TODO
-        print('Something went wrong')
+        print('Could not retrieve Linode: {} error'.format(result))
         exit(2)
 
     potential_matches = potential_matches['data']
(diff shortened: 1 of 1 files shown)


linodecli/plugins/ssh.py
```diff
@@ -0,0 +1,107 @@
...
+        exit(1)
+
+    if exact_match['status'] != 'running':
+        print('{} is not running (status is {}); operation aborted.'.format(
+            label, exact_match['status']))
+        exit(2)
+
+    # find a public IP Address to use
+    public_ip = None
+    for ip in exact_match['ipv4']:
```
displague: A `-6` option would make sense to permit users to get the public IPv6 address instead.

`-6` is convention in many tools, including ssh.

Dorthu: Linodes come with a `/64` for slaac addresses - what of those addresses should we select to SSH to?  The lowest?

displague: Users only get the one address described as the SLAAC address.  The `/64` in this case is not addressable by the user, it is their network of accessible peers that don't require a gateway (`fe80::1`) to reach.  

This is a confusing feature of the API because the IPv4 addresses do not return `/24` and `/17` (for private addresses) which would be inline with the use of `/64` in the SLAAC response. 

In the case of IPv6 Ranges, the entire `/116` is user addressable (across all nodes in the region) and any address outside of that range must use the `fe80::1` gateway to be reached.

patthiel:
Tested this and it works as advertised!

Issue #122, opened 2019-04-17
Explain in the docs how to add multiple tags when creating things


asauber:
You do it like this:

```
asauber@linaret2 ~ $ linode-cli linodes create --tags okay --tags wow --tags great --root_pass 'somethign20343##1'
```

Doesn't appear to be documented anywhere

asauber:
Rather than document this, maybe we should change this arg to use

`nargs='*'`

which has better behavior

gsauthof:
Whatever you do please consider also supporting comma delimited tags, e.g. `--tags tag1,tag2`. Right now this literally generates one tag `tag1,tag2`. I don't think that there is really a use for tags containing a comma, thus, it would make a good delimiter.

lgarber-akamai:
Hey there!

I've put up a [pull request](https://github.com/linode/linode-cli/pull/311) to add documentation for specifying multi-value and nested arguments. I've also created a ticket in our internal tracker to look into implementing alternative methods of specifying lists, although this functionality is not planned to be implemented at the moment.

Let us know if you have any other feedback or suggestions :slightly_smiling_face:

PR #123, opened 2019-04-18
[DNM, breaks interface] Use whitespace separated arguments for arrays

asauber:
Rather than needing to pass the `--flags` argument multiple times like this to pass multiple flags:

`linode-cli linodes create --root_pass someth0343## --tags tag1 --tags tag2 --tags tag2`

This changes the interface to whitespace separated arguments for an array

`linode-cli linodes create --root-pass someth0343## --tags tag1 tag2 tag3`

### DIFF
--- a/linodecli/operation.py
+++ b/linodecli/operation.py
@@ -148,7 +148,7 @@ class CLIOperation:
                 if arg.arg_type == 'array':
                     # special handling for input arrays
                     parser.add_argument('--'+arg.path, metavar=arg.name,
-                                        action='append', type=TYPES[arg.arg_item_type])
+                                        nargs='*', type=TYPES[arg.arg_item_type])
                 else:
                     if arg.arg_type == 'string' and arg.arg_format == 'password':
                         # special case - password input


asauber:
If we combine "append" with "nargs", then the value is a list of lists.

We can flatten, but how do we do that generically in a nice way?

Dorthu:
This is related to #122

Dorthu:
A downside to this approach is that tags allow spaces, but with `nargs='*'` you have to quote the tag that has spaces:

```
linode-cli linodes create --root_pass s3cureP@ssw0rd --tags 'has spaces' other
```

asauber:
I think quoting individual args with spaces is expected behavior.

Dorthu:
This works as expected, and I'm fine with it 👍

Dorthu:
Is there still interest in this?  At this point I'm tempted to close it because it is breaking and not necessary.

asauber:
I've since seen other projects use the multiple-pass version of this so let's close it.

PR #125, opened 2019-05-08
Add SSH Plugin Smoke Tests

patthiel:
* Updated the dockerfile to install the _actual_ version 1.1.0 of bats-core, after much frustration
* Updated the until loops conditionals to the correct comparison form of `[[`
* Added SSH Plugin smoke tests that i believe to be reasonably stable (based on some other tests we do for image packaging 🙃 ).
* Update common utilities to `#!/usr/bin/env bash` instead of `#!/bin/bash` since this will use the correct bash for your env instead of the default that's included with your system (helpful on OSX if you've installed a newer bash and have added that to your path).
* Added `DOCKER_BATS` env var check for when running in a container. 

## To Test:

```shell
# Ensure you have bats-core installed. If not, brew install bats-core should do the trick.

# Uncomment the `.env.example` file and add your CLI Token(s) and save as `.env`

$ cd linode-cli/test
$ bats ssh/distro-and-connection-check.bats
$ bats ssh/ssh.bats
```

PM me to see test results on CI


Dorthu:
Can this be squashed ?

patthiel:
@Dorthu absolutely.

PR #129, opened 2019-06-14
Added `obj` plugin to support Linode Object Storage

Dorthu:
This shouldn't be merged yet :wink:

Adds the `obj` plugin, allowing invocation of Object Storage commands.
Also automatically provisions Object Storage Keys for the account if
none are saved to the config, and helps users request access to the
Early Access Program if they aren't in it yet.

### DIFF
--- a/linodecli/cli.py
+++ b/linodecli/cli.py
@@ -142,6 +142,7 @@ class CLI:
         """
         self.spec = spec
         self.ops = {}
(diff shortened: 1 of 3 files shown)


patthiel:
Worked well in my testing, although one suggestion would be to add columns for the OBJ fields in the CLI. I believe those fields need to be added to the openapi spec in order to show up?

PR #137, opened 2019-09-26
Only suggest an upgrade when an upgrade exists

### LINKED ISSUE #130: Pip2 Fails to install the latest api spec version
When using pip2 to install the linode cli, it installs the correct version, but seems to install the wrong Spec version (4.0.8, instead of the latest 4.1.0). 

This issue has been documented by customers via https://www.linode.com/community/questions/18361/linode-cli-version-mismatch and reproduced. 

I have some theories why this might be happening, and will follow up in this ticket. However, for now, if you encounter this issue, installing the linode cli with `pip3` will fix the issue.

Dorthu:
Closes #130

Right now, when a new version of the Linode API ships, there is some
span of time where the CLI knows it's out of date, but a new version
isn't available.  After some discussion and thought, this change makes
the CLI check PyPI for a newer version before warning that it should be
upgraded.  This check is only performed when the API response's headers
indicate that a new spec version is available.

Also refactored so that all version checking is disabled if
`--suppress-warnings` is used, since it won't matter in that case.

This adds a single additional HTTP request to CLI calls when the API
indicates a new spec version, which is _probably_ okay.  I'm open to
opinions on this.

I tested this by changing the version the CLI build thought it was, and
the spec version it was built off of.  This can't _really_ be tested
until the next API release.

### DIFF
--- a/linodecli/cli.py
+++ b/linodecli/cli.py
@@ -376,37 +376,74 @@ complete -F _linode_cli linode-cli""")
 
         result =  method(url, headers=headers, data=body)
 
-        # if the API indicated it's newer major or minor version than the client, print a warning
-        if 'X-Spec-Version' in result.headers:
-            spec_version = result.headers.get('X-Spec-Version')
+        if not self.suppress_warnings:
+            # check the major/minor version API reported against what we were built
+            # with to see if an upgrade should be available
+            api_version_higher = False
 
-            try:
-                # Parse the spec versions from the API and local CLI.
-                StrictVersion(spec_version)
-                StrictVersion(self.spec_version)
+            if 'X-Spec-Version' in result.headers:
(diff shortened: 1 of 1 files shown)


linodecli/cli.py
```diff
@@ -376,37 +376,74 @@ def do_request(self, operation, args, filter_header=None):
...
+
+                    if pypi_response.status_code == 200:
+                        # we got data back
+                        pypi_version = pypi_response.json()['info']['version']
+
+                        # no need to be fancy; these should always be valid versions
+                        if LooseVersion(pypi_version) > LooseVersion(self.version):
+                            new_version_exists = True
+                except:
+                    # I know, but if anything happens here the end user should still
```
rmcintosh: as long as you know... :p

PR #138, opened 2019-10-03
Check Terraform Version in K8s Plugin

### LINKED ISSUE #131: [k8s-alpha] Exit if terraform < 0.11.14+ or terraform >= 0.12.0 for now
The k8s-alpha plugin should exit if `terraform version` is greater than or equal to 0.12.0 or `terraform version` is less than 0.11.14

https://github.com/linode/linode-cli/blob/master/linodecli/plugins/k8s-alpha.py

rcsalome:
This closes #131

After the Terraform command as been found, we can run `terraform
version` and compare it against our minimum and maximum versions.
Currently, those versions are v0.11.14 and v0.12.0.

The minimum version is an inclusive check, where v0.11.14 is supported.
The maximum version is an exclusive check, where v0.12.0 is not
supported.

It is not clear if Terraform uses true semantic versioning, but we
assume that the "major.minor.patch" are all going to be integers for now.

### DIFF
--- a/linodecli/plugins/k8s-alpha.py
+++ b/linodecli/plugins/k8s-alpha.py
@@ -5,7 +5,7 @@ import argparse
 import base64
 import sys
 import os
-from subprocess import call as spcall, Popen, PIPE
+from subprocess import call as spcall, check_output, Popen, PIPE
 import hashlib
 import shutil
 from terminaltables import SingleTable
@@ -17,6 +17,9 @@ except NameError:
     FileNotFoundError = IOError
 
 plugin_name = os.path.basename(__file__)[:-3]
+
+terraform_min_version = '0.11.14' # This version and above are supported
+terraform_max_version = '0.12.0'  # Only versions below this one are supported
 
 def call(args, context):
     parser = argparse.ArgumentParser("{}".format(plugin_name), add_help=False)
@@ -75,6 +78,13 @@ def create(args, context):
         if 'kubectl' in needed_deps:
             print_kubectl_install_help()
         sys.exit(1)
+    # Check if terraform version is between min and max
+    if not terraform_version_supported(terraform_min_version,terraform_max_version):
+        print('Terraform version unsupported. Must be between v{} and v{}').format(
+                terraform_min_version,terraform_max_version)
+        print_terraform_install_help()
+        sys.exit(1)
+
 
     tf_var_map = create_varmap(context)
 
@@ -200,6 +210,13 @@ def delete(args, context):
             print_terraform_install_help()
         sys.exit(1)
 
+    # Check if terraform version is between min and max
+    if not terraform_version_supported(terraform_min_version,terraform_max_version):
+        print('Terraform version unsupported. Must be between v{} and v{}').format(
+                terraform_min_version,terraform_max_version)
+        print_terraform_install_help()
+        sys.exit(1)
+
     parser = argparse.ArgumentParser("{} create".format(plugin_name), add_help=True)
     parser.add_argument('name', metavar='NAME', type=str,
                         help="The name of the cluster to delete.")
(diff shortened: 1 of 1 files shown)


asauber:
Hi, thanks for implementing this!

Since this issue was created, the new _minimum_ Terraform version for this subcommand is now 0.12.0, as that version made breaking syntax changes with updates found in this commit. https://github.com/linode/terraform-linode-k8s/commit/5827f7d9841941365f3b72c3dc58cb39ed267174

Could you update this PR to reflect that?

linodecli/plugins/k8s-alpha.py
```diff
@@ -281,6 +298,37 @@ def dep_installed(command):
...
+    maximum = [int(i) for i in max_version.split('.')]
+    # Check that version is less than maximum
+    for i in range(len(version)):
+        if version[i] < maximum[i]:
+            return True
+        elif version[i] > maximum[i]:
+            return False
+        else:
+            pass
+    return False
```
Dorthu: You might consider using `LooseVersion` (or `StrictVersion` if you're confident the version numbers will always be in a sane format) to make these checks simpler:

```python
def terraform_version_supported(min_version, max_version):
  version = check_output(['terraform','version']).split()[1].replace('v','')
  return LooseVersion(min_version) < LooseVersion(version) < LooseVersion(max_version)
```

(That example is probably oversimplified, I didn't actually test it)

The CLI already does something similar when checking for updates [here](https://github.com/linode/linode-cli/blob/cd291f6ae3a3c2c0bbbd738a6ab315465a885efd/linodecli/cli.py#L429)

rcsalome:
> Hi, thanks for implementing this!
> 
> Since this issue was created, the new _minimum_ Terraform version for this subcommand is now 0.12.0, as that version made breaking syntax changes with updates found in this commit. [linode/terraform-linode-k8s@5827f7d](https://github.com/linode/terraform-linode-k8s/commit/5827f7d9841941365f3b72c3dc58cb39ed267174)
> 
> Could you update this PR to reflect that?

Sure. I will update the minimum and maximum to 0.12.0 and 0.13.0 respectively.

linodecli/plugins/k8s-alpha.py
rcsalome: Good suggestion! I will definitely make use of the `distutil` library.

Also, I looked a bit more into Terraform versions, and it seems they do stick with semantic versioning. https://github.com/hashicorp/terraform/blob/master/version/version.go#L13-L19

While the release versions are just integers, people can download alpha, beta, rc, etc releases on their own. Here is an example of one that is `Terraform v0.12.0-rc1`. https://releases.hashicorp.com/terraform/0.12.0-rc1/

I'll update the code accordingly.

rcsalome:
@asauber , @Dorthu . I made the recommended changes. They greatly simplified the code. Also, I tested with the `Terraform v0.12.0-rc1` version and it works by treating it as valid between `v0.12.0` and `v0.13.0`. I think that behavior is ok, but it's up to you.

linodecli/plugins/k8s-alpha.py
```diff
@@ -75,6 +79,13 @@ def create(args, context):
         if 'kubectl' in needed_deps:
             print_kubectl_install_help()
         sys.exit(1)
+    # Check if terraform version is between min and max
+    if not terraform_version_supported(terraform_min_version,terraform_max_version):
+        print('Terraform version unsupported. Must be between v{} and v{}').format(
+                terraform_min_version,terraform_max_version)
```
Dorthu: The close paren for the print statement should be after the close paren for `format` here: 

```python
        print('Terraform version unsupported. Must be between v{} and v{}'.format(
                terraform_min_version,terraform_max_version))

```

linodecli/plugins/k8s-alpha.py
```diff
@@ -281,6 +299,11 @@ def dep_installed(command):
     except: 
         return False
 
+def terraform_version_supported(min_version, max_version):
+    # The Terraform version is of the format "Terraform v0.0.0"
+    version = check_output(['terraform','version']).split()[1].replace('v','',1)
```
Dorthu: This breaks in python3.  I think this'll fix it for both 2 and 3:

```suggestion
    version = check_output(['terraform','version']).split()[1].decode().replace('v','',1)
```

rcsalome:
Review comments addressed. Also tested on python2 and python3. Thanks!

asauber:
Tested and prepared to merge after one more change: can you put a single space character after every comma, as seen in the rest of this file and PEP 8?

rcsalome:
No problem. Just added them now.

PR #139, opened 2019-10-09
Fix image tag references

patthiel:
* Fixes tag reference. I think. cc @tnation14

### DIFF
--- a/scripts/cibuild.sh
+++ b/scripts/cibuild.sh
@@ -10,4 +10,4 @@ docker run \
     -e USERID=$(grep Uid /proc/self/status | cut -f2 | awk '{$1=$1};1') \
     -u $(id -u) \
     -v $(pwd):/src \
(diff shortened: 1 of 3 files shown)


scripts/cibuild.sh
```diff
@@ -10,4 +10,4 @@ docker run \
     -e USERID=$(grep Uid /proc/self/status | cut -f2 | awk '{$1=$1};1') \
     -u $(id -u) \
     -v $(pwd):/src \
-    linode-cli-$BUILD_TAG
+    "linode-cli-$BUILD_TAG:latest"
```
tnation14: If the Dockerfile for the CLI builder changes, this line can lead to race conditions between builds. Let's name the image `linode-cli-builder:${BUILD_TAG}` then you can do `docker run linode-cli-builder:${BUILD_TAG}` with some more guarantees that the image you're running is the image you want.

Issue #106, opened 2019-03-04
Cannot filter Events on entity.id


Dorthu:
The entity.id filter for GET /account/events sends in a string instead of an int:

```
linode-cli events list --entity.type linode --entity.id 12928812
Request failed: 400
┌errors────┬──────────────────────────────┐
│ field    │ reason                       │
├──────────┼──────────────────────────────┤
│ X-Filter │ Entity ID must be an integer │
└──────────┴──────────────────────────────┘
```

rcsalome:
I have a small fix for this. Will make a PR.

patthiel:
This has been shipped with Linode-CLI `v2.10.2`! Closing

PR #141, opened 2019-10-16
Add entity.id to argparse for int Conversion

rcsalome:
This is a fix for #106

Args that are not picked up by the argparse are being treated as
strings. A way around this is to add the specific argument,
'--entity.id', to argparse and treat it as an 'int'. We also suppress
the argument when displaying the "help" message.

### DIFF
--- a/linodecli/cli.py
+++ b/linodecli/cli.py
@@ -133,6 +133,7 @@ class CLI:
                     '.'.join(prefix+[name]),
                     info.get('x-linode-filterable') or False,
                     info.get('x-linode-cli-display') or False,
+                    info.get('type') or 'string',
                     color_map=info.get('x-linode-cli-color')))
         return attrs
 
--- a/linodecli/__init__.py
+++ b/linodecli/__init__.py
@@ -70,6 +70,8 @@ def main():
                              "This is useful for scripting the CLI's behavior.")
     parser.add_argument('--version', '-v', action="store_true",
                         help="Prints version information and exits.")
+    # type cast string args to other types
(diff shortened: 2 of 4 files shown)


rcsalome:
Actually, I just realized that I haven't properly tested this. I'll try to get test data in my account to see if the `linode-cli events list --entity.type linode --entity.id <id_number>` command works.

But it does seem that the original 400 error is gone.

rcsalome:
The latest commit message is more accurate.

```
This is a fix for #106

It creates a list "INT_FILTER_ARGS" where any argument added to it will
have an argparse argument created that will treat it as an "int" instead
of a "string".

**Comment by Dorthu** [comment]
I'll have to test this approach, but ideally the CLI doesn't need hardcoded values to designate type; this is generated off of [the spec](https://developers.linode.com/api/docs/v4/openapi.yaml) which should be the authority on the types of arguments.

**Comment by rcsalome** [comment]
Thanks for the tip. I looked at the spec, and you're right. It has the `type` in there. Specifically, it shows that the `events` `list` action returns `entity` of type `object` and those have an `id` of type `integer`.

Here is a more robust approach that uses that `type` for the argparse arguments.

File: linodecli/cli.py [hunk 1]
```diff
@@ -133,6 +133,7 @@ def _parse_properties(self, node, prefix=[]):
                     '.'.join(prefix+[name]),
                     info.get('x-linode-filterable') or False,
                     info.get('x-linode-cli-display') or False,
+                    info.get('type') or 'string',
```
rcsalome: I think this is what is currently happening in a way. So this line preserves the default of `string` as a fallback.

Issue #78, opened 2018-12-06
`make build` fails: cp: cannot stat 'data-2': No such file or directory


themightychris:
On the current master (b49c6302d836d8e64ba0605e8cd7c88cd8eb1969) I get this on `make build`:
```                                                                                                                    
rm -f linodecli/data-*                                                                                                                                       
rm -f linode-cli.sh                                                                                                                                          
rm -f dist/*                                                                                                                                                 
python -m linodecli bake https://developers.linode.com/api/v4/openapi.yaml --skip-config                                                                     
No spec baked.  Please bake by calling this script as follows:                                                                                               
  python3 gen_cli.py bake /path/to/spec                                                                                                                      
Baking...                                                                                                                                                    
Warning: Resolved empty node for {'type': 'object'} in [{'$ref': '#/components/schemas/IPAddress'}, {'type': 'object'}]                                      
Baking bash completions...                                                                                                                                   
Writing file...                                                                                                                                              
Done.                                                                                                                                                        
python3 -m linodecli bake https://developers.linode.com/api/v4/openapi.yaml --skip-config                                                                    
No spec baked.  Please bake by calling this script as follows:                                                                                               
  python3 gen_cli.py bake /path/to/spec                                                                                                                      
Baking...                                                                                                                                                    
Warning: Resolved empty node for {'type': 'object'} in [{'$ref': '#/components/schemas/IPAddress'}, {'type': 'object'}]                                      
Baking bash completions...                                                                                                                                   
Writing file...                                                                                                                                              
Done.                                                                                                                                                        
cp data-2 linodecli/                                                                                                                                         
cp: cannot stat 'data-2': No such file or directory                                                                                                          
make: *** [Makefile:23: build] Error 1    
```

[![asciicast](https://asciinema.org/a/6wZanPJ8oGXAT6ftBjbNgrdqE.svg)](https://asciinema.org/a/6wZanPJ8oGXAT6ftBjbNgrdqE)

Dorthu:
The build process requires both python2 and python3, and expects `python` to be python2.  I've seen this error before if `python --version` returns 3.x.    Try modifying [this line](https://github.com/linode/linode-cli/blob/master/Makefile#L21) of the Makefile to say `python2` instead of `python` - although I'm hesitant to commit this change because many systems don't have a `python2` by default (ideally the Makefile will detect what pythons are available and give reasonable errors if it can't build).

themightychris:
[...]

PR #144, opened 2019-10-31
Edit Makefile to fix issue #78 and allow py2/py3

danieldjewell:
As described in Issue #78, the existing makefile expects "python" to
point to python2. However, on many modern distros "python" actually
points to python3 (and python2 is available by "python2"). This changes
the python2 command to "python2" if the python version detected is
python3. (This *could* create a different issue if python2 isn't
    available with "python2" ... but I *believe* that it should be in
    99-100% of cases?)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,9 +8,11 @@ SPEC ?= https://developers.linode.com/api/docs/v4/openapi.yaml
 ifeq ($(PYTHON), 3)
 	PYCMD=python3
 	PIPCMD=pip3
+	PY2CMD=python2
 else
 	PYCMD=python
 	PIPCMD=pip
+	PY2CMD=python
 endif
 
 install: check-prerequisites requirements build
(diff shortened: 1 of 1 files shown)


patthiel:
Python2 has reached it's end of life. Thinking we should address this now 🙃

alex19EP:
is anyone can comment on that. i want to use linode-cli on arch linux where python is python 3 and python2 is python 2. why cly need two version of python at all?

Dorthu:
Building the CLI includes pickling a python representation of the OpenAPI spec to save time at runtime (it takes several seconds to parse the spec, where loading the pickle is near instant).  Unfortunately, the pickle data format changed between python2 and python3, so in order to ship a universal package, data for both python versions needed to be built.

I will look into this again and see if we have low enough usage of python2 now that we can remove support for it (although unfortunately, despite it being end-of-life, `python` on many systems is still python2).

jriddle-linode:
Fully switched to py3, outdated.

PR #146, opened 2019-11-04
add: clean accounts script

patthiel:
* Adds a script that will clean test accounts prior to running the CLI tests and after.
* Updated test runner arg handling to properly check input arguments, regardless of ordering


patthiel:
> Elegant

this made my morning, thanks d00d. Bash is weird, but fun 🙃

PR #155, opened 2019-11-19
Lke fixes

Dorthu:
This needs https://github.com/linode/linode-api-docs/pull/160

Only respect `--version` if no command is given.  

This broke some LKE commands that use `--version` to supply the version
of kubernetes that should be deployed.  Previously, including
`--version` in invocation would print the version and exit; now it only
does anything if no command is given.  For example:

```bash
linode-cli --version # prints version
linode-cli linodes --version # prints help for linodes command
```

Handle lists of objects 

LKE uses node_pools, a list of objects, to specify what pools must be
supplied when creating an LKE Cluster.  This is the first time we've had
such input, and the CLI didn't handle it well.

This is a slightly scary change as it changes how the spec is parsed -
please test carefully with other endpoints!

### DIFF
--- a/linodecli/__init__.py
+++ b/linodecli/__init__.py
@@ -100,10 +100,16 @@ def main():
         cli.config.set_user(parsed.as_user)
 
     if parsed.version:
-        # print version info and exit
-        print("linode-cli {}".format(VERSION))
-        print("Built off spec version {}".format(cli.spec_version))
-        exit(0)
+        if not parsed.command:
+            # print version info and exit - but only if no command was given
+            print("linode-cli {}".format(VERSION))
+            print("Built off spec version {}".format(cli.spec_version))
+            exit(0)
+        else:
+            # something else might want to parse version
+            # find where it was originally, as it was removed from args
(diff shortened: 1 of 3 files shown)


linodecli/__init__.py
```diff
@@ -100,10 +100,16 @@ def main():
         cli.config.set_user(parsed.as_user)
 
     if parsed.version:
-        # print version info and exit
-        print("linode-cli {}".format(VERSION))
-        print("Built off spec version {}".format(cli.spec_version))
-        exit(0)
+        if not parsed.command:
```
jnschaeffer: This is ugly, but it's basically how I would do it...

Alternatively, you could boot `--version` out of the list of known arguments, then look for it in `args`. Is that better? Is that worse? Who knows?

Dorthu:
fwiw your example invocation would do the same thing without the indexes as this is coded now.  I think it's natural enough to invoke it this way, especially with the established precedent that you define nested items with json path notation, and you specify array items by using the same argument multiple times.

PR #157, opened 2019-11-23
Update: Converts bucket size output to human readable bytes

scottmc87:
Had trouble deploying locally to test this, however I did run this code through multiple tests in an online IDE using various bite sizes and it continually worked successfully. 

This code converts the bytes output of a bucket size `linode-cli obj du` to human readable sizes. (KB, MB, GB, TB etc..) instead of just bytes, for easier to read sizes for our customers.

### DIFF
--- a/linodecli/plugins/obj.py
+++ b/linodecli/plugins/obj.py
@@ -440,6 +440,7 @@ def show_usage(client, args):
             if total > 1024:
                 total = total / 1024 
             if total < 1024:
+                total = round(total, 2)
                 total = str(total) + " " + x
                 break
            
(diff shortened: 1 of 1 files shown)


linodecli/plugins/obj.py
```diff
@@ -432,6 +432,17 @@ def show_usage(client, args):
             total += obj.size
 
         grand_total += total
+
+        # Coverts bucket size to human readable bytes. 
+        total = float(total)
+        denomination = ["KB", "MB","GB","TB"]
+        for x in denomination:
+            if total > 1024:
+                total = total / 1024
```
Dorthu: This leaves `total` as a float, causing it to blow up when applying padding on line 446.  I would just add `str(total)` to that line to make sure it's always a string.

scottmc87: I do convert total to a string in the second condition right before the loop breaks. Would that be sufficient since it's being converted to a string right before it is called to display?

PR #173, opened 2020-03-13
Better error handling for 401 & 403 when enabling Object Storage 

Dorthu:
The first time the `obj` plugin is run, it generates s3 keys for itself
and stores them in the CLI's config.  If that process fails, it spits
out an error - but in all cases, that error was default 401 error the
CLI printed.  That's because plugins previously could use the CLI's
request-handling functionality, but they got the error handling it
included as well.  This change allows the actual response and status
code to get passed back to the plugin (instead of non-200 responses
never making it back to them) and gives the `obj` plugin a nicer error
message in the case that the error was a 401 (which is caused by missing
the required OAuth Scopes for a request).


Dorthu:
Hmm.. good catch.  Looks like right now this attempts to create credentials even if it's only running the help, which is less than ideal. I'll push that up here before merging it.

PR #176, opened 2020-03-25
Handle multipart uploads correctly

### LINKED ISSUE #174: Object Storage put: Error: RequestTimeTooSkewed
Hello, everyone!

During an object storage put, I am receiving `Error: RequestTimeTooSkewed`. How may I leverage timeout?

- The source is a linode server at Freemont, CA
- Object Storage is at Newark, NJ
- File has 7G

Dorthu:
Closes #174

Previously, large files were sent as normal uploads, causing a long
upload time followed by a `EntityTooLarge` error.  This is because s3
has a maximum size for file uploads; this isn't the maximum size for
objects, mind, but for individual uploads.  Files larger than this size
must be uploaded in multiple parts.

This change supports uploading files as multiple parts.  I tested it
with some fabricated large files made with this command:

```bash
dd if=/dev/zero of=file-to-upload bs=1024 count=0 seek=20240
```

### DIFF
--- a/linodecli/plugins/obj.py
+++ b/linodecli/plugins/obj.py
@@ -59,10 +59,10 @@ access, or ask them to generate Object Storage Keys for you."""
 
 
 # Files larger than this need to be uploaded via a multipart upload
-UPLOAD_MAX_FILE_SIZE = 1024 * 1024 * 5
+UPLOAD_MAX_FILE_SIZE = 1024 * 1024 * * 1024 *  5
 # This is how big the chunks of the file that we upload will be
 # This is a float so that division works like we want later
-MULTIPART_UPLOAD_CHUNK_SIZE = 1024 * 1024 * 5.0
+MULTIPART_UPLOAD_CHUNK_SIZE = 1024 * 1024 * 1024 * 5.0
 
 
 def list_objects_or_buckets(client, args):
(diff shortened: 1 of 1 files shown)


linodecli/plugins/obj.py
```diff
@@ -57,6 +58,13 @@
 access, or ask them to generate Object Storage Keys for you."""
 
 
+# Files larger than this need to be uploaded via a multipart upload
+UPLOAD_MAX_FILE_SIZE = 1024 * 1024 * * 1024 *  5
```
patthiel: ` * * ` ?

linodecli/plugins/obj.py
```diff
@@ -190,9 +204,48 @@ def upload_object(client, args):
...
+    print("{} ({} parts)".format(filename, num_chunks))
+
+    try:
+        with open(file_path, "rb") as f:
+            for i in range(num_chunks):
+                print(" Part {}".format(i+1))
+                upload.upload_part_from_file(f, i+1, cb=_progress, num_cb=100, size=MULTIPART_UPLOAD_CHUNK_SIZE)
+    except Exception as e:
+        upload_exception = e
+    finally:
```
patthiel: why does this live in the finally block and not in `except`?

Dorthu: Originally there were other ways it could happen, but I guess they got removed.  I'll clean it up

PR #177, opened 2020-03-25
Allow API POST/PUT Body Elements to Conflict with CLI Flags

### LINKED ISSUE #154: Attempting to create LKE clusters prints CLI version info instead
When attempting to create an LKE cluster with `linode-cli lke cluster-create`, the CLI prints version info instead:
```
$ linode-cli lke cluster-create --label helloworld --version 1.16
linode-cli 2.10.3
Built off spec version 4.9.0
```

It looks like this is because `--version` is specified as a top-level argument, and the CLI prints version info and terminates as soon as it sees `--version`, regardless of whether `command` or `action` were set ([argument definition link](https://github.com/linode/linode-cli/blob/f970c001b68b39ef2b3b746945b830184e1dd807/linodecli/__init__.py#L71-L72), [argument usage link](https://github.com/linode/linode-cli/blob/f970c001b68b39ef2b3b746945b830184e1dd807/linodecli/__init__.py#L102-L106)).

Fixing this in a thorough way would probably involve creating command and action objects corresponding to both what's in the API spec and the commands that are in `__init__.py` currently so that this sort of ambiguous interpretation of arguments between different scopes can't happen.

Dorthu:
Closes #154

For the first time ever, one of the API's request arguments conflicts
with a built-in CLI flag, that being `POST lke/clusters`'s argument
"version".  When translated to the CLI's interface, it becomes
`--version`, which the CLI will pick up as a request to print its own
version number an exit.

This change adds a new spec extension to Linode's OpenAPI spec,
`x-linode-cli-name`, which is allowed in schemas and defines the name
the CLI should use for that element _instead of_ the name of that schema
in the spec.  This should be used only in exceptional cases (like that
noted above) to keep the CLI's behavior predictable.

### DIFF
--- a/linodecli/cli.py
+++ b/linodecli/cli.py
@@ -104,6 +104,7 @@ class CLI:
                 "desc": info.get('description') or '',
                 "name": arg,
                 "format": info.get('format', None),
+                "name_override": info.get('x-linode-cli-name', None),
             }
             # handle input lists
             if args[path]['type'] == 'array' and 'items' in info:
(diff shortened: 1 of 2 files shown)


asauber:
Given that the LKE endpoints are `/beta` endpoints, I would prefer to change the name of the field in the API.

That being said, this logic could be needed in general.

Dorthu:
I'm fine with the API name changing, but I can imagine this coming up again, and it doesn't hurt to have.

Dorthu:
This was already fixed (and more simply) in https://github.com/linode/linode-cli/commit/e371922649d371b9e408e20d0af8a1fbb767b772 - closing this PR.

PR #179, opened 2020-03-27
Allow regenerating Object Storage keys

### LINKED ISSUE #178: No easy way to rotate keys in OBJ plugin
If you use the CLI's obj plugin, it generates you an access key/secret key and stores them in its config.  However, if you should revoke those keys (say through the cloud manager), there's not obvious way to get the plugin back into a working state.

Add a feature to the obj plugin that regenerates keys, and an explanation in the help for the plugin explaining how/why it's needed.

Dorthu:
Closes #178

Previously, there was no way to regenerate your object storage keys
without manually editing the CLI's config, which is a pretty lousy user
experience.  If you were to delete the keys in cloud, for example, the
CLI would be stuck attempting to use them for requests to Object
Storage.

This change adds the "regenerate-keys" command to the `obj` plugin,
allowing object storage keys to be regenerated in case they become
expired or otherwise don't work.

### DIFF
--- a/linodecli/plugins/obj.py
+++ b/linodecli/plugins/obj.py
@@ -585,6 +585,9 @@ def call(args, context):
         tab.inner_heading_row_border=False
         print(tab.table)
         print()
+        print("Additionally, you can regenerate your Object Storage keys using the "
+              "'regenerate-keys' command")
+        print()
         print('See --help for individual commands for more information')
 
         exit(0)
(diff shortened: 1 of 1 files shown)


linodecli/plugins/obj.py
```diff
@@ -613,6 +616,13 @@ def call(args, context):
                 message = e.message
             print('Error: {}'.format(message))
             sys.exit(6)
+    elif parsed.command == "regenerate-keys":
+        print("Regenerating Object Storage keys..")
+        _get_s3_creds(context.client, force=True)
+        print("Done.")
+        print("Warning: Your old Object Storage keys _were not_ expired!  If you want "
```
victoriaroan:
```suggestion
        print("Warning: Your old Object Storage keys _were not_ expired automatically!  If you want "
```
Just to avoid any confusion in case the keys actually were expired and that's why they were changing them.

Dorthu:
Yeah, the `linode-cli obj` commands are part of a plugin.  Plugins are the only part of the CLI that aren't generated off the OpenAPI spec, so we use them to do custom logic (like storing the credentials in the config file) that wouldn't make sense in the spec.  To keep it consistent, plugins are given the CLI object that would be used normally, and can make requests through it as normal, handling the output themselves.

Most Object Storage endpoints are removed from the CLI with a spec extension and are only supported using the plugin, which talks to s3 directly (meaning it's much more efficient and flexible).  However, key management was left in the normal, spec-generated, CLI because it cannot be done directly through Ceph.

victoriaroan:
Ah, interesting. Thanks for the explanation!

PR #195, opened 2020-05-26
Allow invoking obj plugin without config file

Dorthu:
This is for @phillc

Previously, the `obj` plugin would require a config file to use to
get/store its object storage keys.  This isn't great for scripting.
This change allows those keys to be provided via flags to the plugin.

To invoke the plugin without a config, do something like this:

```bash
TOKEN=$API_V4_TOKEN linode-cli obj ls --access-key $ACCESS_KEY --secret-key $SECRET_KEY
```

Should this be pulled from the environment instead of the command line?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -139,6 +139,12 @@ will allow you to set defaults without having to set a token.  Be aware that if
 the environment variable should be unset, the Linode CLI will stop working until
 it is set again or the CLI is reconfigured with a token.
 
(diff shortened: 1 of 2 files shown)


victoriaroan:
I'd vote for environment variables instead of command line flags. Some documentation should be added to the README for both these env vars and the LINODE_CLI_TOKEN var.

Dorthu:
Moved to using the environment instead of flags

PR #198, opened 2020-07-01
Added a default cluster configuration for OBJ

victoriaroan:
Does the following:

- Adds a new OBJ plugin configuration for default cluster
- Removes the default cluster value of 'us-east-1' (to match image/region/type value behavior)
- Adds an OBJ command to configure the plugin (`linode-cli obj configure`)
- Respects `--no-defaults` for cluster value
- If no cluster is specified and no default is set, will ask the user to configure a default (unless `--no-defaults` is used, then just raises an error)
- Allows `--help` to be used for an OBJ command (ex. `linode-cli obj ls --help`) without a cluster and without a valid OBJ key

### DIFF
--- a/linodecli/plugins/obj.py
+++ b/linodecli/plugins/obj.py
@@ -641,6 +641,12 @@ def call(args, context):
 
     parsed, args = parser.parse_known_args(args)
 
+    # don't mind --no-defaults if it's there; the top-level parser already took care of it
+    try:
+        args.remove('--no-defaults')
+    except ValueError:
+        pass
+
     if not parsed.command:
         # show help if invoked with no command
         parser.print_help()
--- a/linodecli/plugins/obj.py
+++ b/linodecli/plugins/obj.py
@@ -62,13 +62,13 @@ access, or ask them to generate Object Storage Keys for you."""
 
 
 # Files larger than this need to be uploaded via a multipart upload
-UPLOAD_MAX_FILE_SIZE = 1024 * 1024 * 1024 *  5
+UPLOAD_MAX_FILE_SIZE = 1024 * 1024 * 1024 * 5
 # This is how big the chunks of the file that we upload will be
 # This is a float so that division works like we want later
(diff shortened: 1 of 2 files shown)


linodecli/plugins/obj.py
```diff
@@ -667,11 +686,39 @@ def call(args, context):
...
+        current_cluster = cluster
+        if current_cluster is None and not context.client.defaults:
+            print("Error: cluster is required.")
+            exit(1)
+
+        if current_cluster is None:
+            print("Error: No default cluster is configured.")
+            print("")
+
+            answer = input_helper("Configure now? [y/N] ")
```
Dorthu, line 708: I understand what you're trying to do, but this is going to break existing scripts that are targeting `us-east-1` assuming it's the default cluster.

As a compromise, you could require a `--cluster` or configuration if there was no stored access key, but if there was an access key in the config preserve the old behavior of defaulting to `us-east-1`.  This would let new users of the obj plugin pick a cluster without having to worry about breaking existing scripts that depend on the current behavior.

Dorthu, line 708: I think you should remove this prompt and just dive into configuring the plugin if necessary - it doesn't work if you say no, and if you don't to pick a cluster you can just `^C`.  This is how the generated CLI handles configuration.

victoriaroan:
While I agree that potentially breaking existing scripts is not ideal, I think bumping the major version number is a good compromise. This PR puts `--cluster` behavior more in line with image/region/type, which makes more sense than having an arbitrary default.

After all, what are major version bumps for if you can't break things every once in a while. 😉

Another option would be to deprecate the original default for a few versions and issue a warning to anyone falling back on it, then remove it with a major version change.

Dorthu:
I don't think if it's worth leaving it in for a few versions - I doubt anyone who's scripted this would see the warnings.  It's an easy fix once it breaks anyway.

PR #202, opened 2020-08-03
Accept passwords from environment

### LINKED ISSUE #200: Secure non-interactive ways for specifying the root password
The linode-cli help and official guides tell me to either enter the root password interactively or on the command line. Thus, for non-interactive uses one has to specify the root password on the command line.

This is suboptimal as specifying the password on the command line is very insecure because any other local user can observe it via `ps`, `top` or by looking into `/proc/...`.

The standard practice to resolve such issues is to let the user alternatively supply a password via a file and/or an environment variable.

Example commands that implement these mechanisms: `virt-builder` and `psql`.

Note that - on Linux - environment variables of a process aren't accessible by other users (except for the root user).

Thus, please add a more secure way to supply the root password when creating linodes in an non-interactive way.

Dorthu:
Closes #200

As it is now, the CLI either accepts passwords on the command line, or
interactively.  The interactive option was added to allow more secure
password entry, but it doesn't cover the automated case as mentioned in
the above issue.

This change implements the suggestion included in #200 to accept
passwords via environment variables to allow secure entry in a
non-interactive setting.  The CLI now gets the values for password
fields like so:

* If a password was given on the command line, use it
* If not, is an environment variable is set for the field, use its value
* If not, prompt for a password

This preserves backwards-compatible behavior while allowing the
environment to be considered.

The CLI will expect environment variables to match the CLI's argument
name, all uppercase, prefixed with `LINODE_CLI_` - for example, the
`--root_pass` argument to `linodes create` will look for
`LINODE_CLI_ROOT_PASS` in the environment.

### DIFF
--- a/linodecli/operation.py
+++ b/linodecli/operation.py
@@ -6,7 +6,7 @@ from __future__ import print_function
 import argparse
 from getpass import getpass
 import json
-from os import path
+from os import path, environ
 
 
 def parse_boolean(value):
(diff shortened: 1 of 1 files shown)


patthiel:
do we want to document this somewhere? Is there a way we can extend the api spec's description of a parameter in the CLI's usage commands?

Dorthu:
Oooh, good call.  I'll open a docs PR for this.

PR #205, opened 2020-08-17
Accept StackScript scripts from file

### LINKED ISSUE #201: Accept script filename when updating or creating it
It's unexpected that I have to use

```
linode-cli stackscripts update --script "$(cat ssh-host-key-stackscript.sh)" 660652
```

instead of the simpler and more convenient

```
linode-cli stackscripts update --script ssh-host-key-stackscript.sh 660652
```

Please also add a way to read a script directly from a file.

I think reading scripts directly from files is genuinely more useful than using the verbatim option value thus perhaps you also want to deprecate the current way. I mean StackScripts are always multi-line, most likely longer than a few lines thus it's getting cumbersome to supply them as option value, fast.

Currently, e.g. the update help reads

```
usage: linode-cli [-h] [--label label] [--description description]
                  [--images images] [--is_public is_public]
                  [--rev_note rev_note] [--script script]
                  stackscriptId

Update StackScript

positional arguments:
  stackscriptId

optional arguments:
  -h, --help            show this help message and exit
  --label label
  --description description
  --images images
  --is_public is_public
  --rev_note rev_note
  --script script
```
i.e. the `--script` option doesn't have a description that points out that the script indeed has to be specified inline. Since the value to `--is_public` is just a non-descriptive `is_public` marker (which doesn't point out which are valid values - perhaps true/false, perhaps 0/1 or even yes/no - we don't know) thus `script` could also be similar non-descriptive and as well be a placeholder for filename!

Dorthu:
Closes #201

Previously, only "ssl-cert" and "ssl-key" formatted fields were
accepted as file paths (and then only if the input looked like a path
and a file existed at it); this change brings that behavior over to
other fields.  The linked docs PR below ties this into the StackScript
"script" field, but it can be used with any field now using a new spec
extension: `x-linode-cli-format`.  As per the included documentation in
this PR, the only valid value is "file" at present; the intention is
that this overrides the "format" given in the OpenAPI spec, but only in
the CLI, and only if present.  Any field with this format will inherit
the above-described path-checking behavior.

I also sorted the spec extension in the README, because they were all out of sorts.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -365,17 +365,17 @@ added to Linode's OpenAPI spec:
 +---------------------+----------+-------------------------------------------------------------------------------------------+
 |x-linode-cli-action  | method   | The action name for operations under this path. If not present, operationId is used.      |
 +---------------------+----------+-------------------------------------------------------------------------------------------+
-|x-linode-cli-command | path     | The command name for operations under this path. If not present, "default" is used.       |
-+---------------------+----------+-------------------------------------------------------------------------------------------+
-|x-linode-cli-skip    | path     | If present and truthy, this method will not be available in the CLI.                      |
-+---------------------+----------+-------------------------------------------------------------------------------------------+
 |x-linode-cli-color   | property | If present, defines key-value pairs of property value: color.  Colors must be understood  |
 |                     |          | by colorclass.Color.  Must include a default.                                             |
++---------------------+----------+-------------------------------------------------------------------------------------------+
+|x-linode-cli-command | path     | The command name for operations under this path. If not present, "default" is used.       |
 +---------------------+----------+-------------------------------------------------------------------------------------------+
 |x-linode-cli-display | property | If truthy, displays this as a column in output.  If a number, determines the ordering     |
 |                     |          | (left to right).                                                                          |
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -363,17 +363,19 @@ added to Linode's OpenAPI spec:
...
 +---------------------+----------+-------------------------------------------------------------------------------------------+
-|x-linode-cli-display | property | If truthy, displays this as a column in output.  If a number, determines the ordering     |
-|                     |          | (left to right).                                                                          |
+|x-linode-cli-action  | method   | The action name for operations under this path. If not present, operationId is used.      |
 +---------------------+----------+-------------------------------------------------------------------------------------------+
 |x-linode-cli-command | path     | The command name for operations under this path. If not present, "default" is used.       |
 +---------------------+----------+-------------------------------------------------------------------------------------------+
-|x-linode-cli-action  | method   | The action name for operations under this path. If not present, operationId is used.      |
+|x-linode-cli-skip    | path     | If present and truthy, this method will not be available in the CLI.                      |
```
victoriaroan, line 370: Shouldn't this be at the end?

PR #206, opened 2020-08-26
Accept nested lists

Dorthu:
The `lke cluster-create` command is unique in the CLI in that it accepts
a list of dicts whose contents may themselves contain a list.  The CLI
didn't have a good way of disambiguating which element of the outer list
you intended inner list elements to belong to.  This made using the
endpoint impossible.

This change allows the spec to specify that an attribute is accepted as
JSON using the `x-linode-cli-format` extension; if the format is set to
"json" this way, the CLI will stop parsing sub-objects or list items and
accept the entire blob as JSON.  This fixes the above issue with the
below-linked docs PR merged.


patthiel:
looks like a good spot for you to add a CLI test ( ͡° ͜ʖ ͡°)

Dorthu:
Gonna leave this open until the docs next release, at which point it will work

PR #209, opened 2020-10-01
Don't send empty lists in requests

Dorthu:
This was pointed out by @adammohammed

This impacts the `object-storage keys-create` operation, and by
extension the `obj` plugin for new users.

This behavior didn't matter before now, which is probably why it went
unnoticed.  The new `bucket_access` argument to `POST /object-storage/keys`
does behave differently when given an empty list, however, and sending
it along caused the generated keys to have no access to anything.

### DIFF
--- a/linodecli/cli.py
+++ b/linodecli/cli.py
@@ -83,11 +83,14 @@ class CLI:
 
         return tmp
 
-    def _parse_args(self, node, prefix=[], args={}):
+    def _parse_args(self, node, prefix=[], args=None):
         """
         Given a node in a requestBody, parses out the properties and returns the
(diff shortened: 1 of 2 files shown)


adammohammed:
I'm still thinking about it myself. 🤔

adammohammed (approved):
Works as expected now! Thank you @Dorthu

victoriaroan (changes requested):
No longer sends an empty list when bucket access is not specified, but I'm still seeing node_pools in the help:
image

victoriaroan (approved):
Nevermind, forgot to make.

Issue #29, opened 2018-06-07
depends on enum34


ryantm:
I was trying to package `linode-cli` for NixOS and I noticed a couple issues with the release artifacts. I'll post a separate issues about them.

The wheel archive metadata.json file says that run_requires `enum34`, but it looks like `enum34` is only targeting a sub-set of the python versions this project is targeting. I'm not that familiar with how wheel archive metadata works, but it would be nice if it could be specified that `enum34` is only necessary for `python2` or whatever.

Dorthu:
Thanks for brining this up, I'll look into it and see what actually requires it.  I believe you're correct in that it's only relevant for python2

cmorgenstern:
I encountered the same issue packaging `linode-cli` for Arch Linux. The end result was a failure to run under python3.7 due to enum34 not being present.

I ended up patching during the build to modify 'requirements.txt` and `setup.py` with the following to get it to work:

`requirements.txt`
```python
terminaltables
colorclass
requests
PyYAML
enum34; python_version < '3.4'
```

`setup.py`
```python
install_requires=["terminaltables","colorclass","requests","PyYAML"],
extras_require={
    ":python_version<'3.4'": ['enum34'],
},
```

lgarber-akamai:
Closing since it looks like this issue is resolved. Feel free to @ me if this issue is still valid.

PR #211, opened 2020-11-17
Changed requirements so that enum34 is only installed on python3.3 an…

atleta:
…d earlier

- as recommended by cmorgenstern in https://github.com/linode/linode-cli/issues/29#issuecomment-476027354

Current dependency on enum34 breaks virtualenvs and (according to thee comments on the above issue) linux packaging for recent python versions.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,4 @@ terminaltables
 colorclass
 requests
 PyYAML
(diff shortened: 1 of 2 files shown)


Dorthu (approved):
Thank you 🙇

PR #216, opened 2020-12-08
Run --help without a config in CLI and obj plugin

### LINKED ISSUE #214: --no-defaults still necessary when using env variable
The #140 issue does not seem fixed for:

`linode-cli obt put x y`
`linode-cli obj put --help`

Then, once I found the `--no-defaults` trick in #140, I got a:

> Error: cluster is required.

And I haven't seen anything about this in the README nor [here](https://www.linode.com/docs/guides/linode-cli/).

### Feedback on User Experience

It feels weird that it's impossible to get the `--help` without having configured the tool.

I'm trying to have a docker container backup my database and I found it a bit difficult to user `linode-cli` without interaction.

1. The help is unavailable without initializing the config
2. With the env variable, there is the #140 issue
3. There is no info about where is the configuration file (found it in the #148 PR file :P )

### Version

3.2.1

It's not the worst problem ever but Linode is pretty good with user experience (in my experience) so I thought you might appreciate the feedback.

Cheers !

Dorthu:
Closes #214

Previously, the CLI Configuration was initialized expecting to load a
config even if `--help` was passed in, which is unintuitive behavior
because the config will prompt for configuration immediately if a file
isn't found.  This was further complicated by the `obj` plugin, which
manages its own configuration and would fail if the CLI wasn't
configured when it was invoked, even with only a `--help` command.

This change is intended to make unconfigured user of the CLI easier by:

 - Allowing `--help` commands to run without configuration
 - Allowing the `obj` plugin to run `--help` commands without
   configuration
 - Providing better errors in the `obj` plugin, namely around cluster,
   to allow users without a config to easily identify and remedy common
   issues.

This should be tested thoroughly, as it touches some fundamental parts
of the CLI's behavior; namely, you should verify that:

 - Configured use works as it does today
 - Calling with `--skip-config` and `--no-defaults` behaves as expected
 - Using environment variables for tokens and obj keys works as expected
 - Incorrectly setup environments for the above produce useful errors

### DIFF
--- a/linodecli/plugins/obj.py
+++ b/linodecli/plugins/obj.py
@@ -785,6 +785,21 @@ def _get_s3_creds(client, force=False):
     if force or access_key is None:
         # this means there are no stored s3 creds for this user - set them up
 
+        # but first - is there actually a config?  If we got this far, creds aren't
+        # being provided by the environment, but if the CLI is running without a
+        # config, we shouldn't generate new keys (or we'd end up doing so with each
+        # request) - instead ask for them to be set up.
+        if client.config.get_value("token") is None:
+            print(
+                "You are running the Linode CLI without a configuration file, but "
+                "object storage keys were not configured.  Please set the following "
+                "variables in your environment: '{}' and '{}'.  If you'd rather ".format(
+                    ENV_ACCESS_KEY_NAME, ENV_SECRET_KEY_NAME
+                )+"configure the CLI, run `linode-cli configure` or unset the "
+                "'LINODE_CLI_TOKEN' environment variable."
+            )
+            exit(1)
+
         # before we do anything, can they do object storage?
         status, resp = client.call_operation('account', 'view')
 
--- a/linodecli/plugins/obj.py
+++ b/linodecli/plugins/obj.py
@@ -683,13 +683,14 @@ def call(args, context):
         os.environ.get(ENV_SECRET_KEY_NAME, None),
     )
 
-    if access_key and not secret_key or secret_key and not access_key:
-        print("You must set both {} and {}, or neither".format(ENV_ACCESS_KEY_NAME, ENV_SECRET_KEY_NAME))
-        exit(1)
+    if not "--help" in args:
+        if access_key and not secret_key or secret_key and not access_key:
+            print("You must set both {} and {}, or neither".format(ENV_ACCESS_KEY_NAME, ENV_SECRET_KEY_NAME))
+            exit(1)
 
-    # not given on command line, so look them up
-    if not access_key:
-        access_key, secret_key = _get_s3_creds(context.client)
+        # not given on command line, so look them up
+        if not access_key:
+            access_key, secret_key = _get_s3_creds(context.client)
 
     cluster = parsed.cluster
     if context.client.defaults:
@@ -707,7 +708,8 @@ def call(args, context):
             exit(1)
 
         if current_cluster is None:
-            print("Error: No default cluster is configured.")
+            print("Error: No default cluster is configured.  Either configure the CLI "
+                  "or invoke with --cluster to specify a cluster.")
             _configure_plugin(context.client)
             current_cluster = context.client.config.plugin_get_value('cluster')
 
--- a/linodecli/__init__.py
+++ b/linodecli/__init__.py
@@ -26,7 +26,7 @@ except:
 BASE_URL = 'https://api.linode.com/v4'
 
 
-cli = CLI(VERSION, BASE_URL, skip_config='--skip-config' in argv)
+cli = CLI(VERSION, BASE_URL, skip_config=('--skip-config' in argv or '--help' in argv))
 
 def main():
     ## Command Handling
--- a/linodecli/configuration.py
+++ b/linodecli/configuration.py
@@ -36,17 +36,18 @@ def input_helper(prompt):
     else:
         return input(prompt)
 
+
 class CLIConfig:
     def __init__(self, base_url, username=None, skip_config=False):
         self.base_url = base_url
         self.username = username
-        self.config = self._get_config()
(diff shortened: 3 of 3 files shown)


victoriaroan (changes requested):
If I call `python3 -m linodecli obj ls` with no config file, `LINODE_CLI_TOKEN` set in env vars, but neither `LINODE_CLI_OBJ_ACCESS_KEY` nor `LINODE_CLI_OBJ_SECRET_KEY` set in env vars, I get an ugly error:

```
python3 -m linodecli obj ls --cluster=
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/configparser.py", line 845, in items
    d.update(self._sections[section])
KeyError: ''

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/vgood/linode-cli/linodecli/__main__.py", line 5, in <module>
    main()
  File "/Users/vgood/linode-cli/linodecli/__init__.py", line 379, in main
    plugins.invoke(parsed.command, plugin_args, context)
  File "/Users/vgood/linode-cli/linodecli/plugins/__init__.py", line 33, in invoke
    plugin.call(args, context)
  File "/Users/vgood/linode-cli/linodecli/plugins/obj.py", line 693, in call
    access_key, secret_key = _get_s3_creds(context.client)
  File "/Users/vgood/linode-cli/linodecli/plugins/obj.py", line 822, in _get_s3_creds
    ['--label', "{}".format(creds_label)])
  File "/Users/vgood/linode-cli/linodecli/cli.py", line 583, in call_operation
    result = self.do_request(operation, args, filter_header=filters, skip_error_handling=True)
  File "/Users/vgood/linode-cli/linodecli/cli.py", line 428, in do_request
    parsed_args = self.config.update(parsed_args)
  File "/Users/vgood/linode-cli/linodecli/configuration.py", line 110, in update
    return self.update_namespace(namespace, dict(self.config.items(username)))
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/configparser.py", line 848, in items
    raise NoSectionError(section)
configparser.NoSectionError: No section: ''
```

Dorthu:
Good catch!  In that case, I decided to have the CLI produce an error and fail, since without a config or keys in the environment, the CLI would regenerate object storage keys with each request and clutter up the account's object storage keys.

linodecli/plugins/obj.py
```diff
@@ -783,6 +785,21 @@ def _get_s3_creds(client, force=False):
...
+        # config, we shouldn't generate new keys (or we'd end up doing so with each
+        # request) - instead ask for them to be set up.
+        if client.config.get_value("token") is None:
+            print(
+                "You are running the Linode CLI without a configuration file, but "
+                "object storage keys were not configured.  Please set the following "
+                "variables in your environment: '{}' and '{}'.  If you'd rather ".format(
+                    ENV_ACCESS_KEY_NAME, ENV_SECRET_KEY_NAME
+                )+"configure the CLI, run `linode-cli configure` or unset the "
+                "'LINODE_CLI_TOKEN' environment variable."
```
victoriaroan, line 799:
```suggestion
                )+"configure the CLI, unset the 'LINODE_CLI_TOKEN' environment "
                "variable and then run `linode-cli configure`."
```

victoriaroan (changes requested):
Hmmm.. I'm not sold on that solution, or perhaps the wording. If I run `linode-cli configure` with `LINODE_CLI_TOKEN` still set as an env var, it uses the token from the env var:
```
Using token from LINODE_TOKEN.
Note that no token will be saved in your configuration file.
    * If you lose or remove LINODE_TOKEN, Linode CLI will stop working.
    * All profiles will use LINODE_TOKEN.
```
(...that should probably be fixed to use the correct token name.)

So if I ran `linode-cli configure` with the token still as an env var, then tried to run whatever command gave me the error you added, I'd still get the error.

At the very least, the last sentence of the error should be changed to say "If you'd rather configure the CLI, unset the 'LINODE_CLI_TOKEN' environment variable and then run `linode-cli configure`."

Yeah, after thinking about it some more, I'm fine with the solution as long as the wording is changed.

---

A couple other semi-related, probably unimportant things I ran into while testing:

1. We treat empty string values differently for `LINODE_CLI_TOKEN` than `LINODE_CLI_OBJ_ACCESS_KEY` and `LINODE_CLI_OBJ_SECRET_KEY`. Running `linode-cli configure` with `LINODE_CLI_TOKEN=""` set as an env var still gives me the "Using token from LINODE_TOKEN" message above. However, if I set one or both of the OBJ env vars to `""`, the plugin treats them the same as not being set at all.
2. We don't let users set their OBJ creds directly with the `linode-cli obj configure` command, which might be useful. Right now they'd have to edit the config file manually.

Dorthu:
The CLI's `obj` support is implemented as a first-party plugin, which allows us to ship proprietary code that isn't based on the spec alongside the spec-generated CLI.  As a matter of principal, the CLI's base configuration function doesn't handle plugin configuration for simplicity, since most users probably won't need it. (the plugin also requires installing an additional library that isn't installed with the CLI).  I could add a `configure` command to the obj plugin to handle setting up its on configuration if you find that useful, but in general it can manage its own config (and does automatically generate/store keys when it doesn't have any).

victoriaroan (approved):

PR #217, opened 2020-12-11
Added linode-cli completion command

### LINKED ISSUE #194: Autocompletion zsh
Hey guys I installed linode-cli on my ubuntu zsh shell computer, but i didint get autocompletion.
How to configure it, or where can i find autocompletion script. 
I cannot find it in /etc/bash_completion.d

Dorthu:
Closes #194

This command prints shell completions for the requested shell to stdout
and exits.  Presently, only bash is supported, but adding support for
other shells should be simple.


patthiel (approved):
works as intended, do we want to advertise this in the readme?

PR #218, opened 2020-12-15
Move to openapi3 for Linode CLI

Dorthu:
This is for us.

This requires:

- [x] Dorthu/openapi3#21
- [x] Dorthu/openapi3#24
- [x] Dorthu/openapi3#25
- [ ] https://github.com/Dorthu/openapi3/pull/52
- [ ] https://github.com/Dorthu/openapi3/pull/53

Currently, the Linode CLI uses an OpenAPI parser that was written to our
OpenAPI spec (at the time the CLI was written) and evolved slowly as
issues were found.  What's resulted is a very organic, free-range parser
that exists specifically for the Linode CLI and is rather difficult to
unwind and maintain.

This change removes the CLI's spec parser in favor of the [openapi3
python package](https://github.com/Dorthu/openapi3), which I wrote after
completing the CLI as a proper spec parser, in part to replace the one
in this project.  The openapi3 package [is already used to validate
Linode's OpenAPI specification during automated builds](https://github.com/linode/linode-api-docs/blob/9da259cbf7f50c3d26fec1f9a227337c6dc87068/.travis.yml#L6),
meaning that transitioning to this parser will ensure the doc's builds
never fall out of sync with the CLI's capabilities.

⚠️ This is a major change, and will require manual testing.

Additional considerations:

- [ ] Ensure all CLI-specific spec extensions are still supported
- [ ] Ensure all first-party plugins function as needed
- [ ] Adjust requirements for installation as necessary
- [ ] How large of a version bump should this be? (in theory it's only a
      refactor, but it also has a lot of potential to break things
      unintentionally).

Notes to self:

 - [ ] The error message behavior in https://github.com/linode/linode-cli/pull/274 should be ported to this branch

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -360,23 +360,27 @@ Specification Extensions
 In order to be more useful, the following `Specification Extensions`_ have been
 added to Linode's OpenAPI spec:
 
-+---------------------+----------+-------------------------------------------------------------------------------------------+
-|Attribute            | Location | Purpose                                                                                   |
-+---------------------+----------+-------------------------------------------------------------------------------------------+
-|x-linode-cli-action  | method   | The action name for operations under this path. If not present, operationId is used.      |
-+---------------------+----------+-------------------------------------------------------------------------------------------+
-|x-linode-cli-color   | property | If present, defines key-value pairs of property value: color.  Colors must be understood  |
-|                     |          | by colorclass.Color.  Must include a default.                                             |
-+---------------------+----------+-------------------------------------------------------------------------------------------+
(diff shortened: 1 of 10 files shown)


patthiel:
need to add the `OpenAPI` package to requirements, as you noted.

regarding the version change, i think given the scope it warrants a minor version bump. There's nothing inherently breaking and it would be confusing for customers to upgrade a major version bump for some under the hood changes.

Dorthu:
This was completed in #423; closing this branch

PR #222, opened 2021-02-12
Fix configuration behavior

Dorthu:
This was pointed out by @caker

When configuring multiple users, the CLI will prompt you to set the user
you're configuring as the active user, used by default when making
requests.  It also skips this question the first time you're configuring
the CLI.  However, confusingly, it still asks even if the user you're
configuring is already the active user, which happens if you only have
one user configured (and are updating their token).

This change removes the prompt to set the newly-configured user as
active if they are already the active user, and also matches the new
language in Cloud Manager for creating a token.

### DIFF
--- a/linodecli/configuration.py
+++ b/linodecli/configuration.py
@@ -281,7 +281,7 @@ Note that no token will be saved in your configuration file.
             print("""
 First, we need a Personal Access Token.  To get one, please visit
 {} and click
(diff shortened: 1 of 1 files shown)


caker:
Looks great - thank you!

victoriaroan (approved):
Looks good, does the expected things.

PR #223, opened 2021-02-13
OAuth-Based Configuration

Dorthu:
When I started using the GitHub CLI, I was very impressed with its OAuth-based configuration in the terminal, so I ported the feature over to the Linode CLI.  This is how a configuration works with it:

```
$ linode-cli configure
Welcome to the Linode CLI.  This will walk you through some
initial setup.

How would you like to configure the CLI?  [W]eb or [t]oken?
'w' for web or 't' for token: w
Your browser will be directed to this URL to authenticate:

https://login.linode.com/oauth/authorize?client_id=e3dc6ddb291b4709ae54&response_type=token&scopes=*

If you are not automatically directed there, please copy/paste the link into your browser
to continue..


Configuring dorthu
```

Not visible above is a browser being opened, directed to the authorization URL, which then redirects back to localhost for the CLI to capture the OAuth response at a server hosted on `localhost:8123` (and serve a simple landing page telling you to return to your terminal).  The entire workflow feels very nice.

I have a few more thoughts:

 * Should this be the default method (i.e. enter for web, "t" for token?)
 * Should `linode-cli configure` accept a `--token` or `--web` to remove that prompt?

The following follow-up is needed:

- [x] Determining an ideal port for the landing page.
- [x] Style the landing page.
- [x] Change Client ID to be that of an official Linode App.
- [x] Test in all supported python versions.
- [x] Test on other OSes (Windows if possible).

### DIFF
--- a/linodecli/configuration.py
+++ b/linodecli/configuration.py
@@ -30,8 +30,7 @@ CONFIG_NAME = 'linode-cli'
 TOKEN_GENERATION_URL='https://cloud.linode.com/profile/tokens'
 
 # This is used for web-based configuration
-# TODO - use an official app
-OAUTH_CLIENT_ID = 'e36db7f4ec4dcffe558f'
+OAUTH_CLIENT_ID = '5823b4627e45411d18e9'
 
 # in the event that we can't load the styled landing page from file, this will
 # do as a landing page
--- a/linodecli/configuration.py
+++ b/linodecli/configuration.py
@@ -55,6 +55,8 @@ class CLIConfig:
 
         self._configured = False
 
+        self.configure_with_pat = "--token" in sys.argv
+
         if not skip_config and not self.config.has_option('DEFAULT', 'default-user') and self.config.has_option('DEFAULT', 'token'):
             self._handle_no_default_user()
 
@@ -372,7 +374,7 @@ r.send();
         url = "https://login.linode.com/oauth/authorize?client_id={}&response_type=token&scopes=*&redirect_uri=http://localhost:{}".format(
             OAUTH_CLIENT_ID, serv.server_address[1]
         )
-        print("""Your browser will be directed to this URL to authenticate:
+        print("""A browser should open directing you to this URL to authenticate:
 
 {}
 
@@ -390,7 +392,8 @@ to continue..
         except KeyboardInterrupt:
             print()
             print("Giving up.  If you couldn't get web authentication to work, please "
-                  "try token auth, and open an issue at github.com/Linode/Linode-CLI")
+                  "try token using a token by invoking with `linode-cli configure --token`, "
+                  "and open an issue at https://github.com/linode/linode-cli")
             sys.exit(1)
 
         return serv.token
@@ -409,8 +412,7 @@ to continue..
         is_default = not self.config.has_option('DEFAULT', 'default-user')
         username = None
 
-        print("""Welcome to the Linode CLI.  This will walk you through some
-initial setup.""")
+        print("""Welcome to the Linode CLI.  This will walk you through some initial setup.""")
 
         if ENV_TOKEN_NAME in os.environ:
             print("""Using token from {env_token_name}.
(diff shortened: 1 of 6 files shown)


stvnjacobs:
> Should this be the default method (i.e. enter for web, "t" for token?)

stvnjacobs:
> Should this be the default method (i.e. enter for web, "t" for token?)

Web should be the default. The `configure` command is run by humans, and this is the most human-friendly way to get the token.

> Should linode-cli configure accept a --token or --web to remove that prompt?

I don't see a need. If I choose `--web` then I am most likely sitting at the computer, ready to click buttons, and would be fine with the prompt. If the intention is to skip the prompt for automated environments, the token can already be passed as an environment variable or written to the config file by provisioning scripts.

> Determining an ideal port for the landing page.

I did want to make a note about this TODO item. You should choose something from the ephemeral range, as this would be for quick one-time use.

https://en.wikipedia.org/wiki/Registered_port

I would also encourage you to check if the port is in use before binding to it, and incrementing it or choosing another at random from the range if it is in use.

Dorthu:
I added a `[ WAIT ]` to the title of this since, at present, our OAuth service doesn't allow variable ports in redirect URIs.  I'm working on a change in that system to allow these when redirecting to `localhost` only, which will allow this change to proceed.

linodecli/configuration.py
```diff
@@ -366,19 +355,32 @@ def do_GET(self):
...
 
+        # figure out the URL to direct the user to and print out the prompt
+        url = "https://login.linode.com/oauth/authorize?client_id={}&response_type=token&scopes=*&redirect_uri=http://localhost:{}".format(
+            OAUTH_CLIENT_ID, serv.server_address[1]
+        )
+        print("""Your browser will be directed to this URL to authenticate:
+
+{}
+
+If you are not automatically directed there, please copy/paste the link into your browser
```
patthiel, line 379: we should add a `[Q]` command to abort this and get back to the original prompt (where a user can then choose `t` for token.

Dorthu: I don't know if we can - the thread is blocked waiting for a request to come into the server.  Ctr+C should still raise a KeboardInterrupt and exit nicely (I handle it below).

Dorthu:
I pushed a change to default configurations to use the web, with a required `--token` argument to `linode-cli configure` to allow for providing a personal access token.  It seems unlikely that users who are using an interactive configuration are not sitting at the terminal with it (for scripted interactions, the `LINODE_CLI_TOKEN` environment variable is recommended).  In the case that you're configuring the CLI on a remote machine or otherwise can't hit the callback server, the prompts make clear how to provide a personal access token, which should avoid confusion.

I added a "Press enter to continue" prompt in place fo the web/token choice because it was very jarring to have a browser open without notice when I used the CLI, especially from an unconfigured state (such as a fresh install).

linodecli/configuration.py
```diff
@@ -4,15 +4,19 @@
 """
 from __future__ import print_function
 
-
 import argparse
+from datetime import datetime
+import re
```
patthiel, line 9:
```suggestion
from re import search
```

linodecli/configuration.py
```diff
@@ -4,15 +4,19 @@
...
+from http import server
+import socket
```
patthiel, line 11:
```suggestion
from socket import gethostname
```

patthiel: etc

patthiel:
just pedantic comments. 

Otherwise, works as expected on osx. I didn't test this functionality on different OS's though (windoze, ubuntu, etc come to mind).

Dorthu:
The necessary upstream change was shipped yesterday; this should be good to go

linodecli/oauth-landing-page.html
```diff
@@ -0,0 +1,46 @@
...
+          M$    M$M:*V       MI*$VVVVV.....*$               
+           VV$             V*...$$VVVV..:FM                 
+           $VVV$        $*:.....V  $$V*V                    
+            VVVVV$   $F:........I                           
+            $VVVVVVM*...........F                           
+             $VVVVV$*.........:FM                           
+               $VVVVV.......*V                              
+                M$VVM....:FM                                
+                  M$$:.*$                                   
+                     $M
```
patthiel, line 35: heheh love this! 👏  @tiffwong

patthiel (approved):
Ran the BATS tests on this branch as well. LGTM :shipit: !

PR #234, opened 2021-04-20
new: Added image-upload first-party plugin

Dorthu:
The new POST /images/upload endpoint allows users to initiate an image
upload and returns a URL to upload the image data to.  While this can be
worked out on the command line, it's much more convenient to have a
plugin that does the entire process, accepting a file path, calling
the new endpoint, and actually handling the upload all in one command.

This new endpoint must be added to the docs before it can be supported
in this plugin.  During the beta period, this endpoint will not be
available to all users, so this plugin offers a reasonable error message
for users who attempt to invoke it and do not yet have access to it.

### DIFF
--- a/linodecli/plugins/image-upload.py
+++ b/linodecli/plugins/image-upload.py
@@ -1,0 +1,131 @@
+"""
+This plugin allows easy uploading of new Images to Linode's Images system.
+
+Usage:
+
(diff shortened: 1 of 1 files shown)


patthiel (approved):
Just requires spec changes 👍

Issue #225, opened 2021-03-09
Add support for wildcard delete all objects in a bucket or folder


jtara1:
Is there a way to deleted all files in a bucket or at least a folder in a bucket? It doesn't seem like it.

I'd like to see support for the following usages:

The wildcard character actually takes all local files in `$PWD` and passes as argument(s) to the command
```
$ linode-cli obj del jtara1.photos *
```

I don't understand why I get this error here. `Q1` is a folder in the bucket containing files.
```
$ linode-cli obj del jtara1.photos Q1/*
S3ResponseError: 403 Forbidden
```

Dorthu:
Thanks for the report!

I did some testing, and I'm also not totally sure why you get a 403 there - I can reproduce it though, so I'll look into it a bit more and report back.

For the time being, you can remove all items in a folder/bucket with a command like this:

```bash
MY_BUCKET="your-bucket-name"
PREFIX="test/" # the trailing slash is necessary
linode-cli obj ls $MY_BUCKET/$PREFIX | tr -s '  ' ' ' | cut  -f4 -d ' ' | xargs -L1 -- linode-cli obj rm $MY_BUCKET/
```

Dorthu:
This should be included in the next CLI release, expected this afternoon

escaper01:
I found this way of deleting a folder 
```bash
linode-cli obj rb $bucket/$folder
```

Dorthu:
The above PR also allowed `linode-cli obj rb --recursive $bucket/$folder` to accomplish this.  Looks like I forgot to close this issue when it shipped, but it's been in the CLI since 5.4.0

PR #237, opened 2021-04-23
Allow recursively deleting not-empty buckets

### LINKED ISSUE #236: Recursive delete of files in a bucket
I've got a few test buckets I've setup with 350 little files.

Unfortunately, linode-cli obj doesn't seem to support recursive delete. It appears I have to use s3cmd, which I can't use because the current released version doesn't work Python 3.9. Can you please add support for: 

- recursive delete, and / or
- the ability to delete a non-empty bucket?

Thanks!

Dorthu:
Closes #236 and #225

A much-requested feature of the obj plugin was to add the ability to
delete buckets that aren't empty, or otherwise remove many objects in
one operation.  Unfortunately this has to happen client-side, so it can
take a while for large buckets.

This change adds a `--recursive` flag to `linode-cli obj rb` that will,
if given, remove all objects from the bucket before removing the bucket.
As implemented here this will not abort in-progress multi-part uploads
or clear stale object versions; that may be added later if it's
necessary (but if such things exist, they will prevent the bucket
deletion from completing successfully).

### DIFF
--- a/linodecli/plugins/obj.py
+++ b/linodecli/plugins/obj.py
@@ -154,9 +154,24 @@ def delete_bucket(get_client, args):
 
     parser.add_argument('name', metavar='NAME', type=str,
                         help="The name of the bucket to remove.")
+    parser.add_argument('--recursive', action="store_true",
+                        help="If given, force removal of non-empty buckets by deleting "
+                             "all objects in the bucket before deleting the bucket.  For "
+                             "large buckets, this may take a while.")
 
     parsed = parser.parse_args(args)
     client = get_client()
+
+    if parsed.recursive:
+        try:
+            bucket = client.get_bucket(parsed.name)
+        except S3ResponseError:
+            print('No bucket named '+parsed.name)
+            sys.exit(2)
+
+        for c in bucket.list():
+            print("delete: {} {}".format(parsed.name, c.key))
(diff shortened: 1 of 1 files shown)


Dorthu:
I'm looking at removing objects using a wildcard (i.e. `linode-cli obj rm my-bucket '*'` or `linode-cli obj rm my-bucket prefix/*`), but using `*` as a wildcard character has the potential to conflict with object keys that are `*`:

```bash
$ linode-cli obj ls my-bucket
2021-04-23 14:26  5  *
```

If i wanted to delete that key today, `linode-cli obj rm '*'` would delete only that key.  If `*` were used as a wildcard character, that operation would accidentally delete everything in the bucket.

I'm open to ideas on how to solve this, as I'd like to do it.

Dorthu:
For context, s3cmd does allow you to accidentially delete everything in your bucket this way: 

```bash
$ s3cmd rm s3://my-bucket/*
delete: 's3://my-bucket/*'
delete: 's3://my-bucket/1'
delete: 's3://my-bucket/10'
delete: 's3://my-bucket/100'
```

patthiel (approved):

escaper01:
I found this way of deleting folder 
```bash
linode-cli obj rb $bucket/$folder
```

PR #238, opened 2021-04-23
Don't require configuration when running `--version`

### LINKED ISSUE #235: Running linode-cli without Linode Personal Token
I have upgraded [`linode-cli` package for NixOS](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/tools/virtualization/linode-cli/default.nix) to latest. But I did run into an issue of linode-cli requiring Linode Personal Token for basic things as showing `version` and `completion bash`. Currently tests are disabled. I need to use these commands without Linode Personal Token, because Hydra (the service that runs CI) won't have a Linode Personal Token. 

I need to run linode-cli without Linode Personal Token to:
1) check built version.
2) generate bash completions

Any idea?

Thanks.

Dorthu:
Closes #235

While `--skip-config` does allow you to find the version without a token
provided, there's no reason to prompt for it at all, and I can see how
this would be disruptive.

This change treats `--version` as `--skip-config` when deciding if the
user should be prompted for configuration if none is present.

### DIFF
--- a/linodecli/__init__.py
+++ b/linodecli/__init__.py
@@ -26,7 +26,10 @@ except:
 BASE_URL = 'https://api.linode.com/v4'
 
 
-cli = CLI(VERSION, BASE_URL, skip_config=('--skip-config' in argv or '--help' in argv))
+# if any of these arguments are given, we don't need to prompt for configuration
+skip_config = any([c in argv for c in ('--skip-config', '--help', '--version')])
+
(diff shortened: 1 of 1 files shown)


patthiel (approved):
Makes sense to me. Also tested this logic and it works as expected 👍

Dorthu:
This should ship this afternoon

PR #245, opened 2021-05-13
fix: Respect --acl-public when doing multipart uploads

Dorthu:
It was brought to my attention that when running a command like

```bash
linode-cli obj put --acl-public /path/to/large-file my-bucket
```

where `large-file` is sufficiently large to be uploaded in multiple
parts, the CLI would not correctly set the ACLs on the resulting object.
This would require a second CLI command to set the ACLs, which is lousy.

This change correctly propagates the intended canned ACLs to the
multipart upload function so that all uploads will receive the expected
ACLs upon completion.


patthiel (approved):
image

R E S P E C T

PR #246, opened 2021-06-03
Started on supporting GET /tags/:label

Dorthu:
And more generally, oneOf/discriminator repsonses


jriddle-linode:
Out of date, denying, may revisit after spec parser refactor.

PR #251, opened 2021-07-13
fix: Check Image upload size before attempting upload

### LINKED ISSUE #250: plugin image-upload should check size before uploading.
The Cloud Manager seems to check size(?) and fails early. Well it fails early anyway : )

But more to the point, the plugin image-upload should check size before uploading.

I use it to upload a 4.3 Gig img.gz that takes 45 minutes only to have it fail with error message file too big!!

Dorthu:
Closes #250

Right now, the `image-upload` plugin just attempts to upload whatever
it's given, but that can waste a lot of time.

This change checks that the compressed size is no larger than the
maximum allowed (5GB) before initiating the upload.

This change does not attempt to catch cases where the _uncompressed_
size of the image is larger than allowed; the maximum uncompressed size
is not constant, and the CLI has no way to know if an Image uploaded
will end up being too large when it gets there.

### DIFF
--- a/linodecli/plugins/image-upload.py
+++ b/linodecli/plugins/image-upload.py
@@ -12,6 +12,7 @@ import os
 from sys import exit
 
 PLUGIN_BASE = 'linode-cli image-upload'
+MAX_UPLOAD_SIZE = 5 * 1024 * 1024 * 1024 # 5GB
 
 
 def _progress(cur, total):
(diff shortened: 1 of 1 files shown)


linodecli/plugins/image-upload.py
```diff
@@ -94,6 +95,11 @@ def call(args, context):
         print("No file at {}; must be a path to a valid file.".format(filepath))
         exit(2)
 
+    # make sure it's not larger than the max upload size
```
patthiel, line 98: Could you move these conditionals up so that it'll complain about them first, before telling the user they haven't inputted a region.. Imagine a scenario where you didn't input a region, then you fix that, then you realize, the file is over 5gb. In that situation, you should have been told before anything else that the file was too big.

alsaihn (approved):

patthiel (approved):

Issue #231, opened 2021-03-23
Not run on Python3.9 


aishlai:
Traceback (most recent call last):
  File "/usr/local/bin/linode-cli", line 5, in <module>
    from linodecli import main
  File "/usr/local/lib/python3.9/site-packages/linodecli/__init__.py", line 14, in <module>
    from .cli import CLI
  File "/usr/local/lib/python3.9/site-packages/linodecli/cli.py", line 16, in <module>
    from .response import ModelAttr, ResponseModel
  File "/usr/local/lib/python3.9/site-packages/linodecli/response.py", line 7, in <module>
    from colorclass import Color
  File "/usr/local/lib/python3.9/site-packages/colorclass/__init__.py", line 11, in <module>
    from colorclass.codes import list_tags  # noqa
  File "/usr/local/lib/python3.9/site-packages/colorclass/codes.py", line 4, in <module>
    from collections import Mapping
ImportError: cannot import name 'Mapping' from 'collections' (/usr/local/lib/python3.9/collections/__init__.py)

Dorthu:
I could not reproduce this in a python3.9 docker container.  What I did was:

```
$ docker run -ti python:3.9 bash
root@35f96f99f449:/# pip install linode-cli
root@35f96f99f449:/# linode-cli linodes list
```

This worked as expected, accepting a token and then listing my Linodes. 

What command is giving you this error?

Could you also include the output of the following commands?

```
pip show colorclass
pip show linode-cli
```

Dorthu:
It looks like there's a patch for the colorclass library that corrects this import, but I'm not completely sure why it's not working in python 3.9 for you, as the deprecation warning says this:

> DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working

The actual commit that removes `Mapping` from `collections` is [here](https://github.com/python/cpython/commit/c47c78b878ff617164b2b94ff711a6103e781753#diff-c92f871cb9b4aa1583c89f82cd416d58a5d70fff41faf41e45dedf775dbe7713), and is only tagged `v3.10.0a6` and `v3.10.0a5` - I don't believe this should be an issue in 3.9.x.

The upstream fix for colorclass is here: https://github.com/Robpol86/colorclass/pull/27

[...]

PR #253, opened 2021-07-16
Remove dependency on colorclass

Dorthu:
Closes https://github.com/linode/linode-cli/issues/231

The colorclass module is not compatible with Python 3.10, and is
currently [archived on github](https://github.com/Robpol86/colorclass)
with issues open about future compatibility issues and no real
suggestion of movement.

This change replaces `colorclass` with the few color codes that were
actually used so far, which removes the dependency entirely and should
allow the CLI to work on python 3.10.

This appears to work where I've tested it, and from what I read it aught
to work the same on Windows too, but I'd like someone to give it a spin
first to be sure.


Dorthu:
This ended up not working on Windows, so it needs more work.

victoriaroan (approved):

PR #257, opened 2021-08-10
bug: Only allow requested defaults to actions

Dorthu:
This should fix the `linode-cli tickets create` action.

The Linode CLI, when being configured, accepts default values for
"region", "image", and "type".  These values were previously added to
every input for a PUT or POST endpoint, and then parsed out as normal.
This had the effect of defaulting those values to the configured
defaults for actions like `linodes create` or `nodebalancers create`
without additional interaction.

However, the `tickets create` action recently began accepting
`--region`, but only for certain types of tickets.  This caused ticket
creation to fail through the CLI if a default region was configured,
unless `--no-defaults` was given, or unless the specific ticket type was
being opened.

This change allows the spec to specify which actions should use which
defaults, allowing this problem to be resolved as well as avoiding it
for future changes to actions.

:warning: This requires a docs update to allow existing actions to
continue using their defaults.  Such a PR will be linked below.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -384,23 +384,26 @@ Specification Extensions
 In order to be more useful, the following `Specification Extensions`_ have been
 added to Linode's OpenAPI spec:
 
-+---------------------+----------+-------------------------------------------------------------------------------------------+
-|Attribute            | Location | Purpose                                                                                   |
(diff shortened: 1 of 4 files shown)


Dorthu:
Required docs PR is here: https://github.com/linode/linode-api-docs/pull/506

PR #278, opened 2021-12-14
bug: Fix SignatureDoesNotMatch when putting files with special characters

Dorthu:
Related to #277

@mechpaul discovered that putting filenames with some special
characters, such as `\` and `$` using the `obj` plugin returns a
SignatureDoesNotMatch error.

To reproduce:

```bash
$ echo "test" > 'test$file\name'
$ python3 -m linodecli obj put 'test$file\name' some-bucket
```

This change properly quotes the URLs such that these characters are
accepted in keys.

### DIFF
--- a/linodecli/plugins/obj.py
+++ b/linodecli/plugins/obj.py
@@ -15,6 +15,12 @@ from terminaltables import SingleTable
 from math import ceil
 from linodecli.configuration import input_helper
 
+try:
+    from urllib.parse import quote_plus
+except ImportError:
+    # python2
+    from urllib import quote_plus
+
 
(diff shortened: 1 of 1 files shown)


mechpaul:
This fix works for uploading, but it causes problems in the web GUI.

image

I was not able to test how it worked for public read. Setting the ACL to Public Read threw another exception: Unable to set object's ACL: SignatureDoesNotMatch.

Should keys in the object bucket be unescaped?

Dorthu:
Hmm.. it does appear that the key is being stored escaped.  That's not what I intended - I'll revisit it.

PR #279, opened 2021-12-17
ci: init github actions

stvnjacobs:
This initializes the github workflow to build the linode-cli wheel. It
does not handle distribution of the package.

It can be triggered at any time by inputting the spec version to build in the Actions workflow and starting the job.

Example run: https://github.com/stvnjacobs/linode-cli/actions/runs/1594022100

Opening as a draft, as there are a few rough spots:
- I like the idea of putting in the spec version, as it allows for some level of control. That being said, it does add some friction. Looking at the release process here, though, it appears that tags are being applied anyway, so there is some manual intervention. We could always pull from the latest release of the docs repo, if we want.
- This does not publish. It's only one more step to do so, but I just wanted to get feedback before adding pubilshing steps.
- It does not use the Makefile or run the tests. The Makefile wants both python interpreters available, but that is not how GitHub actions or other tooling like tox works. The tests require a cacert.pem file which is not part of this repository.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,51 @@
+name: ci
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to build'
+        required: true
+
+jobs:
+  bake:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['2.x', '3.x']
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - run: |
+          pip install --upgrade pip
+          pip install -r requirements.txt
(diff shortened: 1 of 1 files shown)


Dorthu:
I like the idea here, but at present the Linode CLI still supports python2 - the Makefile requires both versions of python so that it can pickle the parsed spec in a way that's compatible with python2 and python3 (since the pickled data format changed between versions).  I'd be happy to take another look at dropping python2 support (honestly it'd make everything a lot simpler), although it probably won't be as easy as just not distributing it anymore - we'd need an upgrade path for current users.  Either way, until that's done, we need both python versions available when we build the CLI for distribution.

stvnjacobs:
For what it is worth, this does build the CLI with pickled data from both python2 and python3. That is not a limitation of GitHub Actions. The comment about the Makefile is regarding the fact that if accepted as-is, there would be two processes for building the distributed wheel. The one used in CI and the ones used in the Makefile. I view having it done two ways as a con. The end result is the same from both processes.

lgarber-akamai:
Closing as the CLI build and release process has been migrated to GHA in #325

PR #284, opened 2022-01-05
bug: Follow redirects when downloading spec

Dorthu:
Recently the openapi.yaml file moved.  Redirects were set up for the old
location, but the Dockerfile for our tests did not follow them, and so
the tests began failing.

This change updated the Dockerfile for the bats tests to follow
redirects when fetching the spec file.


patthiel (approved):
good catch!

PR #287, opened 2022-01-28
Drop Python 2 support

bavedarnow:


Dorthu (changes requested):
This looks good at a glance - will put it through its paces shortly.  Requesting changes because I want to hold off for a few weeks to give existing users time to convert.

PR #289, opened 2022-02-03
Mention Homebrew installation option in the README

tangestani:
I added linode-cli to the Homebrew package manager: Homebrew/homebrew-core#93251

Edited the README to mention that as an installation option.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,10 +12,6 @@ From pypi::
 
    pip3 install linode-cli
 
-From Homebrew::
-
-   brew install linode-cli
-
 From source::
 
    git clone git@github.com:linode/linode-cli.git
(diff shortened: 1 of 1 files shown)


Dorthu:
Thank you for adding the CLI to Homebrew - that's very exciting.  

I'm going to bring this up internally and see how we want to go about endorsing third party distributions of the CLI, as right now the only official distribution channel that we support is installation via pip, and we would likely want to include other endorsed distribution methods in our test plan for new releases to ensure future releases don't cause issues for them.

Dorthu:
After discussing this with the team, it was clear that we absolutely do want to support and feature third party distributions of the CLI.  However, we don't have the resources on this project to commit to testing that each release works through alternate distribution platforms.  How do you feel about moving these notes to a "Community Distributions" section with a note that these distributions are not officially supported and tested with each release?  If you're alright with that, I can make the changes to make it happy.

tangestani:
Sounds good to me. Maybe we can revisit it having it be officially supported in the future.

Dorthu:
Supporting other distribution channels is on our list, but I can't make any promising as to when it might happen.

PR #299, opened 2022-07-18
feat: add examples from OAS to --help prompt

bavedarnow:
This will output the examples provided in the openapi spec to the --help prompt. For example:

```
linodecli linodes create --help 
linode-cli linodes create
Linode Create

Arguments:
  --type: (required) The [Linode Type](/docs/api/linode-types/#types-list) of the Linode you are creating.
...

Example:
  linode-cli linodes create  --type="g6-standard-2" --region="us-east" --image="linode/debian9" --root_pass="aComplexP@ssword" --authorized_keys="['ssh-rsa AAAA_valid_public_ssh_key_123456785== user@their-computer']" --authorized_users="['myUser', 'secondaryUser']" --stackscript_id="10079" --stackscript_data="{'gh_username': 'linode'}" --booted="" --backup_id="1234" --backups_enabled="" --swap_size="512" --label="linode123" --tags="['example tag', 'another example']" --group="Linode-Group" --private_ip="True" --interfaces.label="example-interface" --interfaces.ipam_address="10.0.0.1/24" --interfaces.purpose="vlan"
```

### DIFF
--- a/linodecli/__init__.py
+++ b/linodecli/__init__.py
@@ -559,13 +559,13 @@ def main():
                             arg.description,
                         )
                     )
-                    arg_example = f' --{arg.path}="{arg.example}"'
+                    arg_example = ' --{}="{}"'.format(arg.path, arg.example)
                     examples.append(arg_example)
                 example = "".join(examples)
                 print()
                 print("Example:")
                 print(
-                    f"  linode-cli {parsed.command} {parsed.action} {example}", end=""
+                    "  linode-cli {} {} {}".format(parsed.command, parsed.action, example)
                 )
 
             elif operation.method == "get" and parsed.action == "list":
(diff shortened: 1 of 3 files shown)


Dorthu (changes requested):
This is a fantastic idea, but I'm afraid it needs a little bit of work.

The example in [the API Docs](https://www.linode.com/docs/api/linode-instances/#linode-reboot) for `linode-cli linodes reboot` is this:

```
linode-cli linodes reboot 123
```

The example the CLI shows on this branch is:

```
Example:
  linode-cli linodes reboot  --config_id=""
```

This command _does_ accept a `config_id`, but we shouldn't be showing the empty value in the example like it's required, and likewise the required Linode ID is missing.

I suspect that we should just be using the example from the docs instead of trying to generate one ourselves.  It looks like you added it to the spec parser, so we just need to implement that.

Dorthu:
How do you feel about the approach in https://github.com/linode/linode-cli/pull/303?  That's based off of this branch, but uses the CLI examples in the spec (where they are present).

PR #301, opened 2022-07-18
new users cannot perform common operations until the cli is configured

bavedarnow:
this moves that section higher up in the readme

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -77,6 +77,27 @@ The first time you invoke the CLI, you will be asked to configure (see
 for "region," "image," and "type." If you configure these defaults, you may
 omit them as parameters to actions and the default value will be used.
 
+Configuration
+"""""""""""""
+
+The first time the CLI runs, it will prompt you to configure it.  The CLI defaults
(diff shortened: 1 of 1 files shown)


Dorthu:
The CLI automatically runs the configurator before executing a command if not token is provided/the CLI is not configured, so most users will be able to jump right into common operations and pass through the configuration step automatically.  Using the `linode-cli configure` command is really only needed for advanced users (i.e. multiple user accounts in one CLI).

That said, i'm fine with rearranging the README if you feel strongly about it.

bavedarnow:
> The CLI automatically runs the configurator before executing a command if not token is provided/the CLI is not configured, so most users will be able to jump right into common operations and pass through the configuration step automatically. Using the `linode-cli configure` command is really only needed for advanced users (i.e. multiple user accounts in one CLI).
> 
> That said, i'm fine with rearranging the README if you feel strongly about it.

That makes sense. In my case, I had a config with an invalid token that was preventing me from running the commands. I'm fine with keeping it as-is

PR #303, opened 2022-07-18
add examples from docs

Dorthu:
- feat: add examples from OAS to --help prompt
- Still haven't dropped python2 support yet :(
- Use examples from the API Docs if present


lgarber-akamai:
Closed by #567

Issue #233, opened 2021-04-14
[Suggestion] Support setting default SSH key(s) to be automatically used during creation


Someguy123:
I've looked around the README, `--help`, the issue tracker, and tried just searching "linode-cli default ssh key", but it doesn't appear to be a supported feature.

Currently, it's a pain to use an SSH key with `linode-cli`, because you need to specify the raw key on the command line like so:

```sh
linode-cli linodes create --label test --root_pass 'Hello@World' --authorized_keys 'ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIIdnuvAH4SPWcf8iV5Oj7psh6jKFbN9NH+tcdGYSEfma john@johndoe'
```

It would be great if it was possible to have one or more default SSH keys that would be automatically used during `create` and other commands that accept SSH keys, similar to how you can already select a default region, linode package, and operating system to use. 

For example, in `~/.config/linode-cli` you could specify an SSH public key directly in a config key:

```ini
[DEFAULT]
default-user = johndoe

[johndoe]
token = abcd1234abcd1234
region = eu-central
type = g6-standard-1
image = linode/ubuntu20.04
sshkey = ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIIdnuvAH4SPWcf8iV5Oj7psh6jKFbN9NH+tcdGYSEfma john@johndoe
```

For multiple SSH keys, you could specify `sshkey` multiple times

```ini
sshkey = ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIIdnuvAH4SPWcf8iV5Oj7psh6jKFbN9NH+tcdGYSEfma john@johndoe
sshkey = ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIKSbBdu7wunXL3nlsRrEh6aVsnOfykgCj+GAmnRTPXjs john@iPad
```

Additionally or alternatively, it might be easier to use saved keys on your Linode account

```ini
# Use both of the SSH keys "johndoe" and "iPad" saved on the Linode account
sshkey = johndoe iPad
```

And/or even read them from a file:

```ini
sshkey = ~/.ssh/id_ed25519.pub
sshkey = ~/.ssh/id_rsa.pub
sshkey = /opt/keys/ipad.pub
```

Dorthu:
This is a great idea!

Right now there are a few other ways to set an ssh key.  You can cat the key from a file in your shell: 

```bash
[...]

PR #308, opened 2022-10-17
allowing default for Authorized_Users in Linode:Create

jriddle-linode:
Allows users to configure a user to automatically set for Authorized_Users in Linode:Create, when you provide a --authorized_users field it will overwrite the default

resolves [#233]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,9 +18,9 @@ install: check-prerequisites requirements build
 
 .PHONY: build
 build: clean
-	python -m linodecli bake ${SPEC} --skip-config
+#	python -m linodecli bake ${SPEC} --skip-config
 	python3 -m linodecli bake ${SPEC} --skip-config
-	cp data-2 linodecli/
+#	cp data-2 linodecli/
 	cp data-3 linodecli/
 	$(PYCMD) setup.py bdist_wheel --universal
 
--- a/linodecli/configuration.py
+++ b/linodecli/configuration.py
@@ -569,12 +569,18 @@ Note that no token will be saved in your configuration file.
             "Please select a valid Image, or press Enter to skip",
         )
 
-        config["authorized_keys"] = self._default_thing_input(
-            "Default SSH Key to deploy with new Linodes.",
-            authorized_keys,
-            "Default SSH Key (Optional): ",
-            "Please select a valid SSH Key, or press Enter to skip",
-        )
+        res = self._do_get_request("/profile/sshkeys", token=config["token"])
+        if res['data']:
+            config["use_profile_ssh_keys"] = self._default_thing_input(
+                "If true use the ssh keys added to your Linode Profile for new Linode Instances.",
+                [True,False],
+                "Default Option (Optional): ",
+                "Please select a valid Option, or press Enter to skip",
+            )
+
+        # check if there are ssh keys added to the profile
+        # if so offer the option to default use the profiles ssh keys for linodes
+        # set config["use_ssh_keys"] true/false
 
         # save off the new configuration
         if username != "DEFAULT" and not self.config.has_section(username):
--- a/linodecli/configuration.py
+++ b/linodecli/configuration.py
@@ -545,7 +545,7 @@ Note that no token will be saved in your configuration file.
         regions = [r["id"] for r in self._do_get_request("/regions")["data"]]
         types = [t["id"] for t in self._do_get_request("/linode/types")["data"]]
         images = [i["id"] for i in self._do_get_request("/images")["data"]]
-        sshkeys = [s["label"] for s in self._do_get_request("/profile/sshkeys")["data"]]
+        authorized_keys = [s["label"] for s in self._do_get_request("/profile/sshkeys", token=config["token"])["data"]]
 
         # get the preferred things
         config["region"] = self._default_thing_input(
@@ -569,9 +569,9 @@ Note that no token will be saved in your configuration file.
             "Please select a valid Image, or press Enter to skip",
         )
 
-        config["sshkey"] = self._default_thing_input(
+        config["authorized_keys"] = self._default_thing_input(
             "Default SSH Key to deploy with new Linodes.",
-            sshkeys,
+            authorized_keys,
             "Default SSH Key (Optional): ",
             "Please select a valid SSH Key, or press Enter to skip",
         )
@@ -734,7 +734,7 @@ Note that no token will be saved in your configuration file.
                 self.config.set(username, "region", self.config.get("DEFAULT", "region"))
                 self.config.set(username, "type", self.config.get("DEFAULT", "type"))
                 self.config.set(username, "image", self.config.get("DEFAULT", "image"))
-                self.config.set(username, "sshkey", self.config.get("DEFAULT", "sshkey"))
+                self.config.set(username, "authorized_keys", self.config.get("DEFAULT", "authorized_keys"))
 
                 self.write_config(silent=True)
             else:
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -18,9 +18,9 @@ install: check-prerequisites requirements build
 
 .PHONY: build
 build: clean
-	python -m linodecli bake ${SPEC} --skip-config
+#	python -m linodecli bake ${SPEC} --skip-config
 	python3 -m linodecli bake ${SPEC} --skip-config
-	cp data-2 linodecli/
+#	cp data-2 linodecli/
```
Dorthu, line 23: Was this change intentionally included in this PR?  [It's probably safe to drop python2 support now](https://pypistats.org/packages/linode-cli), but I would figure doing so would warrant its own PR with all the necessary changes.

Dorthu (changes requested):
I'm afraid this doesn't work like you expect.

The "authorized_keys" argument to POST /linode/instances is expected to be [an array of strings](https://www.linode.com/docs/api/linode-instances/#linode-create), each entry being the full ssh key to be installed for the root user.  The [/profile/sshkeys endpoint](https://www.linode.com/docs/api/profile/#ssh-keys-list) is a way to add keys to your account for new Linode deploys, but you deploy them with the "authorized_users" parameter, and it accepts an array of usernames whose keys you want deployed.

For example, on my account I have two ssh keys:

```
wsmith@oth-mpqrg::linode-cli[defaultssh]$ linode-cli sshkeys list --format label,created
┌───────────────┬─────────────────────┐
│ label         │ created             │
├───────────────┼─────────────────────┤
│ work laptop   │ 2018-08-28T13:25:07 │
│ work-laptop-2 │ 2021-09-22T19:11:37 │
└───────────────┴─────────────────────┘
```

If I wanted to deploy all of them to a new LInode, I'd make a request like this:

```
curl https://api.linode.com/v4/linode/instances -X POST \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-type: applicatoin/json" \
  -d '{"region":"us-east", "type":"g6-nanode-1", "image":"linode/debian10", "root_pass":"s3cr3t!", "authorized_users":["wsmith"]}'
```

I checked out this branch, configured it, and rebuilt with a spec that allows this new default for POST /linode/instances, and here's the generated request:

```
wsmith@oth-mpqrg::linode-cli[defaultssh]$ python3 -m linodecli linodes create --root_pass s3cr3t! --debug
> POST https://api.linode.com/v4//linode/instances
> Authorization: [REDACTED]
> Content-Type: application/json
> User-Agent: linode-cli:5.23.0 python/3.8.6
> Body:
>   {"image": "linode/debian10", "root_pass": "s3cr3t!", "authorized_keys": "work laptop", "type": "g6-nanode-1", "region": "us-east"}
>
< HTTP/1.1 400 BAD REQUEST
< Server: nginx
< Date: Tue, 18 Oct 2022 11:34:41 GMT
< Content-Type: application/json
< Content-Length: 77
< Connection: keep-alive
< X-OAuth-Scopes: *
< X-Accepted-OAuth-Scopes: linodes:read_write
< X-Frame-Options: DENY
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Methods: HEAD, GET, OPTIONS, POST, PUT, DELETE
< Access-Control-Allow-Headers: Authorization, Origin, X-Requested-With, Content-Type, Accept, X-Filter
< X-Spec-Version: 4.138.0
< X-Customer-UUID: F4FC4A5C-0999-4F52-9B36FC0FD2E797F5
< X-RateLimit-Limit: 10
< X-RateLimit-Remaining: 9
< X-RateLimit-Reset: 1666092934
< Retry-After: 30
<
Request failed: 400
┌errors───────────┬───────────────────────┐
│ field           │ reason                │
├─────────────────┼───────────────────────┤
│ authorized_keys │ Must be of type Array │
└─────────────────┴───────────────────────┘
```

I think this feature should either allow you to configure what users will get their keys automatically (sending in "authorized_users" as an array of those usernames - restricted users would only be allowed to send their own username), or it should accept a path to a local ssh key file to deploy as default, and should read in that whole file and send its contents (i.e. I might configure it to point to `~/.ssh/id_rsa.pub`, and when sending the value it would read that file and send its contents in the "authorized_keys" array).  In either case, a decision would need to be made about sending the default value if a value was included (since these are both arrays, it could either append the default all the time, or only send the default if no value was given - both would have merits).

jriddle-linode:
I appreciate the full break down, at the moment I'm just trying to understand what does work and how to get the information. I only posted this PR to get an extra eye on the code this is absolutely not the final design.

jriddle-linode:
https://github.com/linode/linode-api-docs/pull/710

linodecli/configuration.py
```diff
@@ -554,6 +566,19 @@ def configure(self):
...
         )
 
+        res = self._do_get_request("/profile/sshkeys", token=config["token"])
+        if 'data' in res:
+            config["authorized_users"] = self._default_thing_input(
+                "Do you want to use the following username as a authorized_user on New Linodes?",
+                [username],
+                "Default Option (Optional): ",
+                "Please select a valid Option, or press Enter to skip",
+            )
```
lgarber-akamai, line 576: It might add some flexibility to allow users to select any number of sub-users as defaults.
Something along the lines of:
- Get a list of users using [this endpoint](https://www.linode.com/docs/api/account/#users-list) 
- Prompt the user to select a user who should be given default access to newly-created Linodes
- If empty input:
  - Break out of the loop
- If user option is specified:
  - Append that user to `authorized_users` 
  - Drop that user from the user options
  - If the list of user options is now empty:
    - Break out of the loop 

Alternatively, we could simplify this to prompt whether the user wants to automatically be granted access to newly-created Linodes. (e.g. `Would you like to grant the current user ({}) access to all newly created Linodes? y/n`)

linodecli/configuration.py
```diff
@@ -145,11 +148,20 @@ def update(self, namespace, allowed_defaults):
...
+            #     default_key: self.config.get(username, default_key)
+            #     for default_key in allowed_defaults
+            #     if self.config.has_option(username, default_key)
+            #     }
+            update_dicts = {}
+            for default_key in allowed_defaults:
+                if not self.config.has_option(username, default_key):
+                    continue
+                value = self.config.get(username, default_key)
+                if default_key == 'authorized_users':
```
lgarber-akamai, line 161: We might want to allow users to specify a list of values for defaults on list fields, either using a string delimiter or using decoded JSON.

lgarber-akamai (changes requested):
Looks good overall! Just a few UX suggestions

linodecli/configuration.py
jriddle-linode: Good catch I meant to clean this up and use that endpoint and totally forgot. I had some issues with writing and reading from the config file with lists so for now i think its best to stick to just one user, maybe we can expand on this in the future to use a list

linodecli/configuration.py
jriddle-linode: as mentioned i had some issues reading and writing the list

lgarber-akamai: Gotcha, no worries. We can keep this functionality and consider supporting multiple entries down the road :+1:

lgarber-akamai (approved):
Tested against https://github.com/linode/linode-api-docs/pull/710 using the following command:

```
linode-cli linodes create --region us-southeast --type g6-standard-1 --image linode/alpine3.16 --root_pass c00lp@ss! --debug
```

PR #313, opened 2022-10-26
feat: Retry OBJ file part uploads on failure

victoriaroan:
If a single part of a multipart upload fails, the whole upload immediately aborts. This PR retries failed parts up to 3 times before aborting.

### DIFF
--- a/linodecli/plugins/obj.py
+++ b/linodecli/plugins/obj.py
@@ -271,7 +271,6 @@ def _do_multipart_upload(bucket, filename, file_path, file_size, policy):
     upload = bucket.initiate_multipart_upload(filename, policy=policy)
 
     num_chunks = int(math.ceil(file_size / MULTIPART_UPLOAD_CHUNK_SIZE))
-    upload_exception = None
 
     print("{} ({} parts)".format(filename, num_chunks))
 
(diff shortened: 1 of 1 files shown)


linodecli/plugins/obj.py
```diff
@@ -275,13 +275,27 @@ def _do_multipart_upload(bucket, filename, file_path, file_size, policy):
...
+    retry_delay = 2
+
     try:
         with open(file_path, "rb") as f:
             for i in range(num_chunks):
                 print(" Part {}".format(i + 1))
-                upload.upload_part_from_file(
-                    f, i + 1, cb=_progress, num_cb=100, size=MULTIPART_UPLOAD_CHUNK_SIZE
-                )
+                for i in range(num_tries):
```
lgarber-akamai, line 285: Just a small change to avoid shadowing the `i` variable defined above.
```suggestion
                for attempt in range(num_tries):
```

lgarber-akamai (changes requested):
Looks good! Just one small thing

linodecli/plugins/obj.py
victoriaroan: oof yeah, good catch!

lgarber-akamai (approved):

victoriaroan:
I tested by adding
```
                        raise S3ResponseError(500, "failed")
```
right before the `upload_part_from_file` call.

I also fixed a `None` being raised instead of the actual exception.

Dorthu (approved):
Looks good 👍 

Tested by adding a failure after the 30th part, and observed the upload fail and clean itself up

PR #315, opened 2022-11-01
linuke-keep safeguard for tests

jriddle-linode:
Allows us to use "linuke-keep" tag on Linode objects to prevent it from being automatically deleted by a test. Useful for testing on a Linode Account that has long term configurations for other work.


lgarber-akamai (approved):
Looks good! Tested locally against an instance tagged with `linuke-keep`.

Issue #149, opened 2019-11-08
[enhancement/suggestion] Command aliases to match standard *NIX commands


danieldjewell:
I had a thought/suggestion: As it stands now, the commands in linode-cli are functional but, IMHO, are less than intuitive. For example, *nix admins are completely familiar with using commands like ```ls; cp; rm; mv```. I think it would be more user friendly and intuitive to provide aliases (which seems like a good idea instead of flat out modifying the existing command syntax and disrupting users who already have memorized linode-cli commands). 

In other words, provide "ls" as an alias for "list" (and so on) so that one could run: 

```linode-cli linodes ls```

Additionally, supporting a form of auto-completion similar to what is present on many routers and switches would also be helpful. In other words, one would only need to type enough of a command so as to make it unique: there are two "L" commands - "linodes" and "longview" ... so "li" and "lo" are enough to infer the meaning. (This is similar to routers and switches where something like ```sh ip int br``` is shorthand for ```show ip interfaces brief```.) 

The combination of these two things would make it possible to do, for example: 

```linode-cli li ls``` 

in order to list all linodes. If one then aliases "linode-cli" to something shorter (i.e. ```alias lc='linode-cli'```) ... one could shorten this up to a very nice compact command: ```lc li ls```. 

Just a thought - Docker already utilizes the ls/rm/cp syntax... And I think some of the Azure CLI stuff does too? Not suggesting that just because others do something that makes it the "best" or "right" way...

jtara1:
to summarize:
- make aliases for core commands like: `ls`, `cp`, `rm`, `mv`, etc
- autocomplete (linux, macos, windows)

PR #316, opened 2022-11-02
new: Add support for action aliasing

lgarber-akamai:
This change adds support for aliasing actions by defining multiple values for the `x-linode-cli-action` in the API spec.

For example, for GET `/linode/instances`:

```
x-linode-cli-action:
  - list
  - ls
```

would allow users to run either

```
linode-cli linodes ls
```

or 

```
linode-cli linodes list
```

Additionally, all aliases will be displayed on the help page as shown below:

image

See #149, https://github.com/linode/linode-api-docs/pull/708

### DIFF
--- a/linodecli/__init__.py
+++ b/linodecli/__init__.py
@@ -519,7 +519,8 @@ def main():
             print("linode-cli {} [ACTION]".format(parsed.command))
             print()
             print("Available actions: ")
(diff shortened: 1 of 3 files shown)


jriddle-linode (approved):
best change to cli since i added an alias to it in my bashrc file `cli='linode-cli'`

Issue #58, opened 2018-10-11
Show warning when using default values


Dorthu:
More than one person has been confused when trying to create an empty linode or disk and being prompted for a root password.  This behavior comes from the default usage of "image" from the config in requests where it is relevant.  While this behavior is often desirable, it can be confusing.

Add a warning that mentions the default usage and it's value, and how to disable defaults.

PR #318, opened 2022-11-04
output: print default values being used

jriddle-linode:
- warn users when using defaults which values are being used
- warning can be suppressed with `--no-headers` and `--suppress-warnings`
- added missing artifact to make clean

example

resolves [#58]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,4 +26,5 @@ check-prerequisites:
 clean:
 	rm -f linodecli/data-*
 	rm -f linode-cli.sh
-	rm -f dist/*
+	rm -f data-*
+	rm -rf dist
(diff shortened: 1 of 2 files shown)


lgarber-akamai:
@jriddle-linode Tested locally and the logic works great, but I think the warning message could be improved a bit here:

image

It might be a bit more user-friendly to format the warning using the keys of the default values:

```
Using default values for the following fields: `region`, `image`, `type`. These defaults can be disabled by specifying the `--no-defaults` flag.
```

jriddle-linode:
The only issue i have with that is not every response posts the values used (authorized_users for example), let's try it like this for a bit and if we feel it clutters up too much we can change it.

PR #320, opened 2022-11-08
fix: Install `boto` in OCI image Dockerfile

lgarber-akamai:
This change installs the `boto` package during the OCI image build process. This is necessary for running the `obj` plugin.

### DIFF
--- a/Dockerfile-release
+++ b/Dockerfile-release
@@ -16,7 +16,7 @@ FROM python:3.9-slim-buster
 
 COPY --from=builder /src/dist /dist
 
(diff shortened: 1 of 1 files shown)


jriddle-linode (approved):
Would this be better in requirements.txt?

LBGarber:
> Would this be better in requirements.txt?

Hmm, I was reluctant to add this dep to `requirements.txt` since it seems to have been explicitly excluded from there for the `obj` module. I think the idea is to reduce bloat for the vast number of users who don't use the OBJ plugin.

PR #324, opened 2022-11-28
new: add page-size flag

jriddle-linode:
## 📝 Description

Allow users to set page-size field and limit pages between 25 and 500.

## ✔️ How to Test

```bash
make install
linode-cli stackscripts list --page-size 500
```


lgarber-akamai (approved):
Nice! Tested locally against various resources and pagination seems to work properly under all tested conditions.

PR #327, opened 2022-12-01
fix: Check for unrestricted access during configuration step

lgarber-akamai:
## 📝 Description

This change checks whether the token specified has unrestricted access before attempting to access restricted endpoints during the configuration step.


Dorthu (approved):
Tested by configuring an unrestricted and restricted user on this branch - both worked as expected  👍

richardmcintosh (approved):

PR #343, opened 2023-01-10
moving shell completion to its own module

jriddle-linode:
## 📝 Description

Moving all code related to Shell Completions to its own module and adding unit tests

## ✔️ How to Test

`make test`

### DIFF
--- a/linodecli/completion.py
+++ b/linodecli/completion.py
@@ -24,17 +24,20 @@ def get_completions(ops, help_flag, action):
     Handle shell completions based on `linode-cli completion ____`
     """
     if help_flag or not action:
-        return """linode-cli completion [SHELL]
-
-        Prints shell completions for the requested shell to stdout.
-        Currently, only completions for bash and fish are available."""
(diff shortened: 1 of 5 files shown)


linodecli/completion.py
```diff
@@ -0,0 +1,114 @@
...
+    with open("linode-cli.sh", "w", encoding="utf-8") as bash_f:
+        print("Writing file...")
+        bash_f.write(rendered)
+
+def get_completions(ops, help_flag, action):
+    """
+    Handle shell completions based on `linode-cli completion ____`
+    """
+    if help_flag or not action:
+        return """linode-cli completion [SHELL]
```
zliang-akamai, line 27: Use triple quotes here will make the `str` inconsistent with the one before refactoring. Maybe consider use multiple `str` with newline character?

```python
return (
    "str1\n",
    "str2"
)
```

lgarber-akamai (approved):
Nice work!

PR #344, opened 2023-01-17
refactoring configuration and auth

jriddle-linode:
## 📝 Description

Refactoring code related to managing the .linode-cli configuration file
Additional refactoring and tests will be done in other PRs to break up the work some

### DIFF
--- a/linodecli/configuration/auth.py
+++ b/linodecli/configuration/auth.py
@@ -1,5 +1,5 @@
 """
-Helper functions for configuration related to web
+Helper functions for configuration related to auth
 """
 
 import os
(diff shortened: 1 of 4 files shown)


linodecli/configuration/auth.py
```diff
@@ -0,0 +1,224 @@
...
+<h2>Success</h2><br/><p>You may return to your terminal to continue..</p>
+<script>
+// this is gross, sorry
+let r = new XMLHttpRequest('http://localhost:{port}');
+r.open('GET', '/token/'+window.location.hash.substr(1));
+r.send();
+</script>
+"""
+
+@staticmethod
```
zliang-akamai, line 29: Static method decorator may not be needed as this function is not in a class scope.

lgarber-akamai (approved):
Looks good to me other than @zliang-akamai suggestion; assuming we'll be adding tests in a separate PR 🙂

PR #345, opened 2023-01-18
ref: Break out API request logic

lgarber-akamai:
## 📝 Description

This change refactors the API request logic and outdated version handling into a new `api_request.py` module.

Additionally, this change introduces a new test populator framework for mocking various CLI classes (CLI, CLIOperation, etc.)

## ✔️ How to Test

`make test`


jriddle-linode (approved):

zliang-akamai:
Am reviewing it

zliang-akamai (approved):
Looks good! 
BTW, for the tests module, have we decided to use `unittest`? Some people would prefer `pytest` because of more features provided and so it is more popular in modern projects, but `pytest` [can be working with unittest](https://docs.pytest.org/en/7.2.x/how-to/unittest.html#unittest). I think both are good options.

zliang-akamai (approved):
Nice work!

PR #348, opened 2023-01-24
ref: refactoring and unit testing config

jriddle-linode:
## 📝 Description

Refactor all code relevant to generating the linode-cli config file
Add unit tests to the config generation

## ✔️ How to Test

`make test`

### DIFF
--- a/linodecli/configuration/__init__.py
+++ b/linodecli/configuration/__init__.py
@@ -377,8 +377,6 @@ If you prefer to supply a Personal Access Token, use `linode-cli configure --tok
 
         for k, v in config.items():
             if v:
-                print(k)
-                print(v)
                 self.config.set(username, k, v)
 
         self.write_config()
--- a/linodecli/configuration/__init__.py
+++ b/linodecli/configuration/__init__.py
@@ -377,6 +377,8 @@ If you prefer to supply a Personal Access Token, use `linode-cli configure --tok
 
         for k, v in config.items():
             if v:
+                print(k)
+                print(v)
                 self.config.set(username, k, v)
 
         self.write_config()
--- a/linodecli/configuration/__init__.py
+++ b/linodecli/configuration/__init__.py
@@ -290,7 +290,7 @@ Note that no token will be saved in your configuration file.
             token = os.getenv(ENV_TOKEN_NAME)
 
         else:
-            if _check_browsers():
+            if _check_browsers() and not self.configure_with_pat:
                 print("""
 The CLI will use its web-based authentication to log you in.
 If you prefer to supply a Personal Access Token, use `linode-cli configure --token`.
--- a/linodecli/configuration/__init__.py
+++ b/linodecli/configuration/__init__.py
@@ -83,7 +83,7 @@ class CLIConfig:
         """
         if self.default_username() == username:
             print(
-                f"Cannot remote {username} as they are the default user! You can "
+                f"Cannot remove {username} as they are the default user! You can "
                 "change the default user with: `linode-cli set-user USERNAME`"
             )
             sys.exit(1)
--- a/linodecli/configuration/__init__.py
+++ b/linodecli/configuration/__init__.py
@@ -199,62 +199,51 @@ class CLIConfig:
 
         return self.config.get(username, full_key)
 
-    def update_namespace(self, namespace, new_dict):
-        """
-        In order to update the namespace, we need to turn it into a dict, modify it there,
-        then reconstruct it with the exploded dict.
-        """
-        ns_dict = vars(namespace)
-        warn_dict = {}
-        for k in new_dict:
-            if k.startswith("plugin-"):
-                # plugins set config options that start with 'plugin-' - these don't
-                # get included in the updated namespace
-                continue
-            if k in ns_dict and isinstance(k, list):
-                ns_dict[k].append(new_dict[k])
-            if k in ns_dict and ns_dict[k] is None:
-                warn_dict[k] = new_dict[k]
-                ns_dict[k] = new_dict[k]
-        if not any(x in ["--suppress-warnings", "--no-headers"] for x in sys.argv):
-            print(
-                f"using default values: {warn_dict}, "
-                "use --no-defaults flag to disable defaults"
-            )
-        return argparse.Namespace(**ns_dict)
-
+    # TODO: this is more of an argparsing function than it is a config function
+    # might be better to move this to argparsing during refactor and just have
+    # configuration return defaults or keys or something
     def update(self, namespace, allowed_defaults):
         """
         This updates a Namespace (as returned by ArgumentParser) with config values
         if they aren't present in the Namespace already.
         """
         if self.used_env_token and self.config is None:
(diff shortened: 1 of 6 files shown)


tests/configuration.py
```diff
@@ -0,0 +1,280 @@
...
+        Test CLIConfig.configure() with
+        no default user, no environment variables, and no browser
+        """
+        conf = configuration.CLIConfig(self.base_url, skip_config=True)
+
+        def mock_input(prompt):
+            answers = (a for a in ["1", "1", "1", "1"])
+            if 'token' in prompt.lower():
+                return "test-token"
+            return next(answers)
```
lgarber-akamai, line 228: Clever use of iterators!

jriddle-linode: shoutout to some guy on stackoverflow

linodecli/configuration/__init__.py
```diff
@@ -275,7 +199,53 @@ def plugin_get_value(self, key):
...
+            if self.config.has_option(username, key):
+                value = self.config.get(username, key)
+            else:
+                value = allowed_defaults[key]
+            if key == "authorized_users":
+                ns_dict[key] = [value]
+                warn_dict[key] = [value]
+            else:
+                ns_dict[key] = value
+                warn_dict[key] = value
```
lgarber-akamai, line 241: Wonder if we could break this key/value normalization logic into a helper function

linodecli/configuration/helpers.py
```diff
@@ -45,6 +61,30 @@ def _get_config(load=True):
...
+        and not KNOWN_GOOD_BROWSERS.intersection(webbrowser._tryorder)
+    ):
+        print("""
+This tool defaults to web-based authentication,
+however no known-working browsers were found.""")
+        while True:
+            r = input("Try it anyway? [y/N]: ")
+            if r.lower() in "yn ":
+                result = r.lower() == "y"
+                break
```
lgarber-akamai, line 85:
```suggestion
                return r.lower() == "y"
```

linodecli/configuration/helpers.py
```diff
@@ -45,6 +61,30 @@ def _get_config(load=True):
...
     return conf
 
+def _check_browsers():
+    # let's see if we _can_ use web
+    result = True
+    try:
+        webbrowser.get()
+    except webbrowser.Error:
+        # there are no browsers installed
+        result = False
```
lgarber-akamai, line 71: I think we can return false here and drop `result` from the function altogether.
```suggestion
        return False
```

lgarber-akamai (changes requested):

tests/configuration.py
```diff
@@ -0,0 +1,280 @@
...
+                contextlib.redirect_stdout(io.StringIO()),
+                patch.dict(os.environ, {}), requests_mock.Mocker() as m):
+            m.get(f'{self.base_url}/profile', json= {"username": "cli-dev"})
+            m.get(f'{self.base_url}/profile/grants', status_code=204)
+            m.get(f'{self.base_url}/regions', json= {"data":[{"id": "test-region"}]})
+            m.get(f'{self.base_url}/linode/types', json= {"data":[{"id": "test-type"}]})
+            m.get(f'{self.base_url}/images', json= {"data":[{"id": "test-image"}]})
+            m.get(f'{self.base_url}/account/users',
+                  json= {"data":[{"username": "cli-dev", "ssh_keys": "testkey"}]})
+            conf.configure()
```
lgarber-akamai, line 241: Getting this error here; thinking we just need an additional mock statement:
```
requests_mock.exceptions.NoMockAddress: No mock address: POST https://linode-test.com/profile/tokens
```

jriddle-linode: weird it shouldn't be calling that url, i guess i need to make sure its not actually going to go to browser

linodecli/configuration/__init__.py
jriddle-linode: i feel since it relies so heavily on self.config and effectively returns 2 dictionaries its fine as it is now but if we refactor it into argparse i think we can clean it up some

tests/configuration.py
lgarber-akamai: For reference it's being called from here: https://github.com/linode/linode-cli/blob/8e7c69f174a34c9f132ea7fcfe606cd2ed637c3c/linodecli/configuration/auth.py#L124

Wonder why it's using web auth :thinking:

jriddle-linode: i added a mock to force it to false that should fix it

zliang-akamai:
Does pytest run the unit test cases in this PR?

jriddle-linode:
> Does pytest run the unit test cases in this PR?

no it uses unittest, you can run `make test` or check it for the specific command

tests/configuration.py
lgarber-akamai: Looks like your patch wasn't quite working because the current patch references `_check_browsers`  in the `helpers` package rather than the `configuration` package where it is imported. This is because it effectively copies the function ref when using `from ... import ...`. Just a bit of Python jankiness lol

This should fix it up :slightly_smiling_face: 

```python
patch('linodecli.configuration._check_browsers', lambda: False),
```

jriddle-linode: updated good catch

lgarber-akamai (approved):
Tests are passing now, nice work!

linodecli/configuration/__init__.py
```diff
@@ -326,67 +277,36 @@ def configure(
...
-                username, config["token"] = _get_token_terminal(self.base_url)
-            else:
-                # pylint: disable=line-too-long
-                print()
-                print(
-                    "The CLI will use its web-based authentication to log you in.  "
-                    "If you prefer to supply a Personal Access Token, use `linode-cli configure --token`. "
-                )
-                print()
+            if _check_browsers():
```
lgarber-akamai, line 293: Looks like `self.configure_with_pat` was accidentally excluded here.

Maybe something like 
```python
if _check_browsers() and not self.configure_with_pat:
```

lgarber-akamai (changes requested):
Oops, missed something

linodecli/configuration/__init__.py
```diff
@@ -286,31 +203,65 @@ def plugin_get_value(self, key):
...
+            or allowed_defaults is None):
+            return namespace
+
+        warn_dict = {}
+        ns_dict = vars(namespace)
+        for key in allowed_defaults:
+            if key not in ns_dict:
+                continue
+            if ns_dict[key] is not None:
+                continue
```
zliang-akamai, line 231: Just double check, is this intent to continue when `key` is in `ns_dict` and the `value` is `None`?

jriddle-linode: yes the ns_dict is the command arguments passed with the command, if it exists in ns_dict then we don't want to overwrite the custom setting with the default.

linodecli/configuration/__init__.py
jriddle-linode: good catch

zliang-akamai (approved):

lgarber-akamai (approved):

PR #349, opened 2023-01-25
new: Refactor and add unit testing to output-related modules

lgarber-akamai:
## 📝 Description

This pull request refactors and adds unit tests for all output and response-related logic and classes.

## ✔️ How to Test

```
make test
```

### DIFF
--- a/linodecli/output.py
+++ b/linodecli/output.py
@@ -197,6 +197,7 @@ class OutputHandler:  # pylint: disable=too-few-public-methods
         if self.headers and header is not None:
             content = [header]
 
+        # We're not using model attrs here
         if isinstance(columns[0], str):
             return content + data
 
--- a/linodecli/output.py
+++ b/linodecli/output.py
@@ -59,14 +59,17 @@ class OutputHandler:  # pylint: disable=too-few-public-methods
         else:
             header = columns
 
-        if self.mode == OutputMode.table:
-            self._table_output(header, data, columns, title, to)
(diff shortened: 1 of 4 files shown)


lgarber-akamai:
I'd like some feedback as to whether we should refactor the `OutputHandler.print(...)` function to return the data to be output rather than printing it directly. I'm leaning towards leaving it the same since `OutputHandler` implies it also handles outputting/printing data.

linodecli/output.py
```diff
@@ -173,29 +167,48 @@ def _select_json_elements(self, keys, json_res):
...
+            header=None,
+            value_transform=lambda attr, model: model):
+        """
+        Returns the `content` to be displayed by the corresponding output function.
+        `value_transform` allows functions to specify how each value should be formatted.
+        """
+
         content = []
 
+        if self.headers and header is not None:
```
zliang-akamai, line 204: I think `self.headers` is almost equivalent to `self.headers and header is not None` here unless the default behavior is overridden by the magic method. Is that the case?

zliang-akamai:
Sorry, that was an incorrect comment, deleted.

zliang-akamai (approved):

jriddle-linode (approved):
Maybe down the line we can make this a return and print class but I'm not sure how A. that would impact plugins if they use output handler and B. from my perspective it just makes unit testing stuff easier but we already have unit tests so

PR #358, opened 2023-02-09
First Refactoring of obj Plug-In

zliang-akamai:
Type annotations, simplified logics, simplified functions and function calls

### DIFF
--- a/linodecli/plugins/obj.py
+++ b/linodecli/plugins/obj.py
@@ -775,6 +775,9 @@ COMMAND_MAP = {
 
 
 def print_help(parser: ArgumentParser):
+    """
+    Print out the help info to the standard output.
+    """
     parser.print_help()
 
     # additional help
@@ -799,6 +802,9 @@ def print_help(parser: ArgumentParser):
 
 
 def get_obj_args_parser():
+    """
+    Initialize and return the argument parser for the obj plug-in.
+    """
     parser = ArgumentParser(PLUGIN_BASE, add_help=False)
     parser.add_argument(
         "command",
--- a/linodecli/plugins/obj.py
+++ b/linodecli/plugins/obj.py
@@ -773,6 +773,7 @@ COMMAND_MAP = {
     #'info': get_info,
 }
 
+
 def print_help(parser: ArgumentParser):
     parser.print_help()
 
@@ -814,7 +815,10 @@ def get_obj_args_parser():
     )
     return parser
 
-def call(args: List[str], context: PluginContext):  # pylint: disable=too-many-branches,too-many-statements
+
+def call(
+    args: List[str], context: PluginContext
+):  # pylint: disable=too-many-branches,too-many-statements
     """
     This is called when the plugin is invoked
     """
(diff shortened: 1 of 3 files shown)


linodecli/plugins/obj.py
```diff
@@ -798,50 +819,46 @@ def call(args, context):  # pylint: disable=too-many-branches,too-many-statement
...
-    parsed, args = parser.parse_known_args(args)
 
-    # don't mind --no-defaults if it's there; the top-level parser already took care of it
-    try:
-        args.remove("--no-defaults")
-    except ValueError:
-        pass
+def call(
+    args: List[str], context: PluginContext
+):  # pylint: disable=too-many-branches,too-many-statements
```
jriddle-linode, line 827: i would like to remove these pylint disables when possible, i threw them up when adding lint so we could ignore the errors while we refactored, but now hopefully our refactoring fixes these errors and they don't need to be disabled

zliang-akamai: Thank you for the suggestion! Let me try if the linter is still reporting errors. If the errors are still there, may I keep this disables and fix them in the next PR? I would like to keep each PR small and easier for reviews.

zliang-akamai: Unfortunately, it is still reporting the errors...

linodecli/plugins/obj.py
```diff
@@ -798,50 +819,46 @@ def call(args, context):  # pylint: disable=too-many-branches,too-many-statement
...
+    """
+    This is called when the plugin is invoked
+    """
+    if not HAS_BOTO:
+        # we can't do anything - ask for an install
+        pip_version = "pip3" if sys.version[0] == 3 else "pip"
```
lgarber-akamai, line 833: Wondering if we can drop this expression now that Python 2 support is dropped.

linodecli/plugins/obj.py
```diff
@@ -798,50 +819,46 @@ def call(args, context):  # pylint: disable=too-many-branches,too-many-statement
...
     access_key, secret_key = (
-        os.environ.get(ENV_ACCESS_KEY_NAME, None),
-        os.environ.get(ENV_SECRET_KEY_NAME, None),
+        os.getenv(ENV_ACCESS_KEY_NAME, None),
+        os.getenv(ENV_SECRET_KEY_NAME, None),
     )
 
     if not "--help" in args:
-        if access_key and not secret_key or secret_key and not access_key:
+        if bool(access_key) != bool(secret_key):
```
lgarber-akamai, line 861: I love how this logic cleans up :tada:

lgarber-akamai (approved):
Clean!

linodecli/plugins/obj.py
zliang-akamai: Yea, that's on my TODO list.

linodecli/plugins/obj.py
zliang-akamai: Thank you! ^_^

jriddle-linode (approved):

PR #359, opened 2023-02-09
Move consistent fixtures to conftest.py

zliang-akamai:
Putting global fixtures in `conftest.py` to allow pytest to automatically collect and use the fixtures without making manual imports.


ezilber-akamai (approved):

lgarber-akamai (approved):
Nice, always great to clean up import statements!
Tested locally and all tests are passing.

PR #362, opened 2023-02-10
Fix Landing Page Path

zliang-akamai:
We have a landing page in `linodecli/oauth-landing-page.html` but it was never used due to an incorrect path.
This PR fixes this issue and display it instead of the hard coded default landing page in the source code.

Another thing is that we can use Python's [`pathlib`](https://docs.python.org/3/library/pathlib.html) to replace the lower-level operating system APIs `os.path` for files and path manipulations in most situations, which makes it simplify and easy to understand.

### DIFF
--- a/linodecli/configuration/auth.py
+++ b/linodecli/configuration/auth.py
@@ -2,7 +2,6 @@
 Helper functions for configuration related to auth
 """
 
-import os
 import re
 import socket
 import sys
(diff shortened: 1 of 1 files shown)


ezilber-akamai (approved):

lgarber-akamai (approved):
Tested locally and everything works as intended. Nice work!

PR #368, opened 2023-02-15
ref: argument parsing

jriddle-linode:
## 📝 Description

refactor argument parsing code, main function is too big and impossible to navigate
adding unit tests for some of the argument parsing code

## ✔️ How to Test

`make test`


linodecli/__init__.py
```diff
@@ -6,21 +6,24 @@
...
+    help_with_ops,
+    register_args,
+    register_plugin,
+    remove_plugin,
+)
 from .cli import CLI
 from .completion import bake_completions, get_completions
-from .configuration import ENV_TOKEN_NAME
 from .helpers import handle_url_overrides
-from .operation import CLIArg, CLIOperation, URLParam
```
zliang-akamai, line 23: Can you undo these deletions? I they may be imported somewhere else in some in-progress work.

Generally, I think we should be careful about the deletions of "unused" imports within the package (deleting unused imports from third party package is less risky, like `requests` and `yaml` in this case) in `__init__.py`.

lgarber-akamai (approved):
Looks good to me other than @zliang-akamai's change

tests/unit/test_arg_helpers.py
```diff
@@ -0,0 +1,256 @@
...
+        assert "test summary" in captured.out
+        assert "API Documentation" in captured.out
+        assert "https://website.com/endpoint" in captured.out
+        assert "Arguments" not in captured.out
+        assert "filter results" in captured.out
+        assert "filtername" in captured.out
+
+    def test_bake_command_bad_website(self, capsys, mocker, mock_cli):
+        with pytest.raises(SystemExit) as ex:
+            arg_helpers.bake_command(mock_cli, "https://website.com")
```
lgarber-akamai, line 228: Are we using mocking here?

jriddle-linode: Ah good catch, i was using it when i wrote it originally but realized i can just not mock it since the bad use of the function gives me the correct systemexit response anyway.

zliang-akamai (approved):

PR #394, opened 2023-03-08
migrate: bats test to pytest

ykim-1:
This PR migrates the existing bats test to pytest

## ✔️ How to Test
Local:
To run individual test suite. e.g. `python -m pytest tests/integration/domains/test_domain_records.py`
To run all integration tests. e.g. `make testint`

PR #397, opened 2023-03-13
new: Add support for `/acctest` command

lgarber-akamai:
## 📝 Description

This change adds support for the `/acctest` command, which allows maintainers to run integration tests on PRs using a command and commit hash. This implementation mimics the implementation for other DX-owned repos (terraform-provider-linode, ansible_linode, etc.)

The following environment variables will need to be added to the secrets of this repository:
- DX_ACCTEST_APP_ID
- DX_ACCTEST_PRIV_KEY

**NOTE: This change excludes bats tests due to the planned removal of bats in the near future. See #394**

## ✔️ How to Test

```
/acctest sha=mycommitsha
```

```
/acctest sha=mycommitsha tests=test_plugin_image_upload.py::test_file_upload
```

See https://github.com/lgarber-akamai/linode-cli/pull/1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@
 # Makefile for more convenient building of the Linode CLI and its baked content
 #
 
+INTEGRATION_TEST_PATH :=
+
 SPEC_VERSION ?= latest
 ifndef SPEC
 override SPEC = $(shell ./resolve_spec_url ${SPEC_VERSION})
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -48,7 +50,7 @@ testunit:
 
 .PHONY: testint
 testint:
-	pytest tests/integration
+	pytest tests/integration/${INTEGRATION_TEST_PATH}
```
zliang-akamai, line 53: Can it still run full test suite when `INTEGRATION_TEST_PATH` isn't set?

zliang-akamai (approved):

Makefile
lgarber-akamai: Yep! I checked this locally and it seems to work as expected

PR #399, opened 2023-03-13
Added warning when not using `--json` on `linodes ips-list`

ezilber-akamai:
## 📝 Description

Added warning when not using `--json` on `linodes ips-list`

## ✔️ How to Test

`make testunit`

### DIFF
--- a/linodecli/cli.py
+++ b/linodecli/cli.py
@@ -491,8 +491,8 @@ class CLI:  # pylint: disable=too-many-instance-attributes
             and "--json" not in args
         ):
             print(
-                "Not using --json mode for ips-list will result in unreadable output." + 
-                " Use --json instead. A fix is on our roadmap but is not yet implemented.",
+                "Not using --json mode for ips-list will result in unreadable output."
+                + " Use --json instead. A fix is on our roadmap but is not yet implemented.",
                 file=sys.stderr,
             )
 
--- a/linodecli/cli.py
+++ b/linodecli/cli.py
@@ -491,7 +491,8 @@ class CLI:  # pylint: disable=too-many-instance-attributes
             and "--json" not in args
         ):
             print(
-                "Not using --json mode for ips-list will result in unreadable output. Use --json instead. A fix is on our roadmap but is not yet implemented.",
+                "Not using --json mode for ips-list will result in unreadable output." + 
+                " Use --json instead. A fix is on our roadmap but is not yet implemented.",
                 file=sys.stderr,
             )
 
--- a/linodecli/cli.py
+++ b/linodecli/cli.py
@@ -485,6 +485,16 @@ class CLI:  # pylint: disable=too-many-instance-attributes
         action
         """
 
+        if (
+            command == "linodes"
+            and action == "ips-list"
+            and "--json" not in args
+        ):
+            print(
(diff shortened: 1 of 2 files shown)


lgarber-akamai:
Great work so far! I did a bit of digging and it turns out there are a few other commands that have this issue (`firewalls rules-list`, etc.)

To account for these, we might be able to check whether the depth of the displayed content > 3 in `output.py`.

For example:

```python
    @staticmethod
    def _get_data_depth(data, current_depth=0):
        """
        Returns the depth of a dict contains a combintation of dicts and lists.
        """
        
        if not isinstance(data, (dict, list)):
            return current_depth

        items = data.values() if isinstance(data, dict) else data

        if len(items) < 1:
            return current_depth

        return max([OutputHandler._get_data_depth(v, current_depth + 1) for v in items])


    def _attempt_print_display_warning(self, raw_data, response_model):
        if len(raw_data) < 1 or not isinstance(raw_data[0], dict):
            return

        data = raw_data[0]

        # Get all of the top-level column names to be displayed
        columns = {v.name.split(".")[0] for v in self._get_columns(response_model)}

        # Get only the fields to be displayed
        data_filtered = {k: v for k, v in data.items() if k in columns}

        if self._get_data_depth(data_filtered) > 3:
            print(
                "This output contains a nested structure that may not properly be displayed by linode-cli.",
                "A fix is currently on the roadmap but has not yet been implemented.",
                "Please use --json for endpoints like this in the meantime.",
                file=sys.stderr
            )
```

I haven't thoroughly tested this implementation, but it might be a good starting point 👍

linodecli/cli.py
```diff
@@ -485,6 +485,18 @@ def handle_command(self, command, action, args):
         action
         """
 
+        if (
+            (command == "linodes" and action == "ips-list")
+            or (command == "firewalls" and action == "rules-list")
```
lgarber-akamai, line 490: If you want to use some Python hackery, this could become something like:

```python
if (command, action) in [("linodes", "ips-list"), ("firewalls", "rules-list")]:
```

This is more of a stylistic change so feel free to ignore it 🙂

lgarber-akamai (approved):

lgarber-akamai:
> > Great work so far! I did a bit of digging and it turns out there are a few other commands that have this issue (`firewalls rules-list`, etc.)
> > To account for these, we might be able to check whether the depth of the displayed content > 3 in `output.py`.
> > I haven't thoroughly tested this implementation, so it might need some tweaks and unit tests for edge cases 👍
> 
> I would also prefer a less static way of handling it to keep the spirit of the generated CLI we should avoid hard setting as much as possible.

I definitely agree with you there, although since this is a temporary band-aid solution I think it might be good to avoid the additional overhead of more complicated warning logic while we work on a proper solution.

linodecli/cli.py
ezilber-akamai: That's much cleaner. Thanks!

ezilber-akamai:
I agree with Lena here. I think that since this isn't meant to be a long-term fix it would be better to use a simple solution that is easy to expand upon if the bug is discovered in other commands. Should we open this for discussion with the team?

zliang-akamai (approved):

PR #405, opened 2023-03-16
ref: Remove usage of deprecated `distutils` version system

lgarber-akamai:
## 📝 Description

This pull request removes the dependency on the deprecated `distutils` version parser.

## ✔️ How to Test

```
make testunit
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 terminaltables
 requests
 PyYAML
-packaging==23.0
+packaging>=23.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 terminaltables
 requests
 PyYAML
+packaging==23.0
--- a/linodecli/api_request.py
+++ b/linodecli/api_request.py
@@ -4,14 +4,11 @@ This module is responsible for handling HTTP requests to the Linode API.
 
 import json
 import sys
-from distutils.version import (  # pylint: disable=deprecated-module
-    LooseVersion,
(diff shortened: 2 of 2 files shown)


yec-akamai (approved):

requirements.txt
```diff
@@ -1,3 +1,4 @@
 terminaltables
 requests
 PyYAML
+packaging==23.0
```
zliang-akamai, line 4:
```suggestion
packaging>=23.0
```

I think >= might be better since someone might want to use a specific version of `packaging` in the future.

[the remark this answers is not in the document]
zliang-akamai: Yes, `==` is to pin the dependency version number.

zliang-akamai (approved):

requirements.txt
```diff
@@ -1,3 +1,4 @@
 terminaltables
 requests
 PyYAML
+packaging>=23.0
```
okoriko, line 4: @lgarber-akamai when installed through pip with `pip install linode-cli` this new dependency doesn't seem to be pulled. Isn't it necessary to add it here too e.g. [here](https://github.com/linode/linode-cli/blob/dev/setup.py#L78)?

lgarber-akamai: @okoriko Good catch! I'll put up a quick patch to fix this

lgarber-akamai: @okoriko This issue has been resolved in linode-cli `v5.34.1` :tada:

PR #407, opened 2023-03-20
Added `get-kubeconfig` plugin

### LINKED ISSUE #347: [Feature]: allow to merge kubeconfig of lke cluster using cli without having to do that manually
### Description

Currently, after creation of lke cluster using linode-cli, one has to download and process cluster's kubeconfig as described in https://github.com/linode/linode-cli/issues/247. 

Suggest to have an option to merge cluster's kubeconfig to ~/.kube/config similar to https://docs.aws.amazon.com/cli/latest/reference/eks/update-kubeconfig.html 

### Potential Usage Example

`linode-cli lke kubeconfig-view [cluster_id] --merge-to-kubeconfig`

ezilber-akamai:
## 📝 Description

Adds a new `linode-cli get-kubeconfig ...` command that allows users to merge a kubeconfig from an existing cluster to their local machine.

## ✔️ How to Test

`make tetsunit`

Resolves #347


jriddle-linode (approved):
Don't forget to link/close the issue.

linodecli/plugins/get-kubeconfig.py
```diff
@@ -0,0 +1,170 @@
...
+        base64.b64decode(kubeconfig["kubeconfig"]).decode()
+    )
+    currentConfig = None
+
+    if os.path.exists(os.path.expanduser(parsed.kubeconfig)):
+        currentConfig = _load_config(os.path.expanduser(parsed.kubeconfig))
+
+    # If there is no current kubeconfig, dump the cluster config to the specified file location.
+    # If there is a current kubeconfig, merge it with the cluster's kubeconfig
+    if not currentConfig:
```
lgarber-akamai, line 84: This could be simplified to something like:
```python
clusterConfig = _merge_dict(currentConfig, clusterConfig) if currentConfig is not None else clusterConfig

if parsed.dry_run:
    ....
else:
    ....
```

linodecli/plugins/get-kubeconfig.py
```diff
@@ -0,0 +1,170 @@
...
+    os.makedirs(os.path.dirname(filepath), exist_ok=True)
+    with open(filepath, "w", encoding="utf-8") as file_descriptor:
+        yaml.dump(data, file_descriptor)
+
+
+# Merges the lists in the provided dicts. If non-list properties of the two
+# dicts differ, uses the value from the first dict.
+def _merge_dict(dict_1, dict_2):
+    for key in dict_1:
+        if isinstance(dict_1[key], list):
```
lgarber-akamai, line 154: We can use guard clauses here:
```python
if not isinstance(dict_1[key], list):
    continue
```

tests/unit/test_plugin_kubeconfig.py
```diff
@@ -0,0 +1,233 @@
...
+    assert "Error retrieving kubeconfig:" in stderr_buf.getvalue()
+
+
+# Test the output when the provided path exists but the file there contains improper YAML
+def test_improper_file(mock_cli):
+    stderr_buf = io.StringIO()
+    mock_cli.call_operation = mock_call_operation
+
+    os.makedirs(os.path.expanduser("~/fake/path"), exist_ok=True)
+    with open(os.path.expanduser("~/fake/path/config"), "w", encoding="utf-8"):
```
lgarber-akamai, line 125: We might be able to use TemporaryFiles here to avoid accidentally overwriting real files: https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile

lgarber-akamai (changes requested):
Nice work! Just a few stylistic/testing improvements

tests/unit/test_plugin_kubeconfig.py
```diff
@@ -0,0 +1,233 @@
...
+
+    assert result["name"] == yaml_a["name"]
+    assert result["things"] == yaml_a["things"]
+    assert result["items"] == yaml_a["items"] + [yaml_b["items"][1]]
+    assert result["dictionary"] == yaml_a["dictionary"]
+
+
+@pytest.fixture(scope="session", autouse=True)
+def fake_kubeconfig_file():
+    with tempfile.NamedTemporaryFile(delete=False) as fp:
```
lgarber-akamai, line 205: Nice!

linodecli/plugins/get-kubeconfig.py
```diff
@@ -0,0 +1,168 @@
...
+        # Add all dict_2 entries with unique names to new_entries
+        for index, value in enumerate(dict2_names):
+            if value not in dict1_names:
+                new_entries.append(dict_2[key][index])
+
+        # Concat dict_1 list with new_entries and update dict_1 list to concatenated list
+        merged = dict_1[key] + new_entries
+        dict_1[key] = merged
+
+    return dict_1
```
lgarber-akamai, line 168: We can optimize this a bit using a map to merge values by name:
```python
def _merge_dict(dict_1, dict_2):
    # Return a new dict to prevent any accidental mutations
    result = {}

    for key in dict_1:
        if not isinstance(dict_1[key], list):
            result[key] = dict_1[key]
            continue

        merge_map = {
            sub["name"]: sub for sub in dict_1[key]
        }

        for sub in dict_2[key]:
            # If the name is already in the merge map, skip
            if sub["name"] in merge_map:
                continue

            merge_map[sub["name"]] = sub

        # Convert back to a list
        result[key] = [v for v in merge_map.values()]

    return result
```

lgarber-akamai (changes requested):
Great work! Tests are passing locally; just one optimization to the merge logic and we should be good to go

linodecli/plugins/get-kubeconfig.py
ezilber-akamai: Nice! Thanks for the tip

linodecli/plugins/get-kubeconfig.py
```diff
@@ -0,0 +1,169 @@
...
+    if not data:
+        print(f"Could not load file at {filepath}", file=sys.stderr)
+        sys.exit(1)
+
+    return data
+
+
+# Dumps data to a yaml file
+def _dump_config(filepath, data):
+    os.makedirs(os.path.dirname(filepath), exist_ok=True)
```
zliang-akamai, line 141: Can you use `pathlib` high-level APIs here (and elsewhere where `os.path` is used)? I think it's more readable and simple.

linodecli/plugins/get-kubeconfig.py
```diff
@@ -0,0 +1,169 @@
...
+
+    # If --label was used, fetch the kubeconfig using the provided label
+    elif parsed.label:
+        kubeconfig = _get_kubeconfig_by_label(parsed.label, context.client)
+    else:
+        print("Either --label or --id must be used.", file=sys.stderr)
+        sys.exit(1)
+
+    # Load the specified cluster's kubeconfig and the current kubeconfig
+    clusterConfig = yaml.safe_load(
```
zliang-akamai, line 74: Can you change the var name to be Python style?
https://peps.python.org/pep-0008/#function-and-variable-names

linodecli/plugins/get-kubeconfig.py
zliang-akamai: https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

zliang-akamai: You can convert str to be Path object where it enters the code. For this line specific, it would just be something like `Path(filepath).parent.mkdir(parents=True, exist_ok=True)`. I think this is a more OOP way instead of utilizing the low-level operating system's API.

ezilber-akamai: Got it, that worked. Thanks!

linodecli/plugins/get-kubeconfig.py
```diff
@@ -0,0 +1,169 @@
...
+    if not data:
+        print(f"Could not load file at {filepath}", file=sys.stderr)
+        sys.exit(1)
+
+    return data
+
+
+# Dumps data to a yaml file
+def _dump_config(filepath, data):
+    Path.mkdir(PurePath.parent(Path(filepath)), exist_ok=True)
```
lgarber-akamai, line 141: IIRC `PurePath.parent` is a property method, so we can simplify this to:
```python
Path.mkdir(filepath.parent, exist_ok=True)
```

lgarber-akamai (changes requested):
One last change and we should be good to go!

linodecli/plugins/get-kubeconfig.py
```diff
@@ -0,0 +1,169 @@
...
+    # If there is a current kubeconfig, merge it with the cluster's kubeconfig
+    cluster_config = (
+        _merge_dict(current_config, cluster_config)
+        if current_config is not None
+        else cluster_config
+    )
+    if parsed.dry_run:
+        print(cluster_config)
+    else:
+        _dump_config(Path(parsed.kubeconfig).expanduser(), cluster_config)
```
zliang-akamai, line 92: Can you save a path object at about line 78 and use that object instead of parsing from `str` every time?
Something like:
```python
kubeconfig_path = Path(parsed.kubeconfig).expanduser()
```

zliang-akamai:
```suggestion
    kubeconfig_path = Path(parsed.kubeconfig).expanduser()
    if Path(parsed.kubeconfig).expanduser().exists():
        current_config = _load_config(kubeconfig_path)

    # If there is no current kubeconfig, dump the cluster config to the specified file location.
    # If there is a current kubeconfig, merge it with the cluster's kubeconfig
    cluster_config = (
        _merge_dict(current_config, cluster_config)
        if current_config is not None
        else cluster_config
    )
    if parsed.dry_run:
        print(cluster_config)
    else:
        _dump_config(kubeconfig_path, cluster_config)
```

zliang-akamai (approved):

PR #416, opened 2023-03-22
setup: fix required deps mismatch

chenrui333:
## 📝 Description

Fix missing packaging dependency in the setup.py file

relates to https://github.com/Homebrew/homebrew-core/pull/126192

## ✔️ How to Test

Tested out in the homebrew CI.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,6 +65,9 @@ else:
     version = get_version()
     bake_version(version)
 
+with open('requirements.txt') as f:
+    requirements = f.read().splitlines()
+
 setup(
     name="linode-cli",
     version=version,
@@ -75,11 +78,7 @@ setup(
     url="https://www.linode.com/docs/api/",
     packages=find_packages(include=['linodecli*']),
     license="BSD 3-Clause License",
-    install_requires=[
-        "terminaltables",
-        "requests",
-        "PyYAML",
-    ],
+    install_requires=requirements,
     entry_points={
         "console_scripts": [
             "linode-cli = linodecli:main",


chenrui333:
cc @lgarber-akamai @zliang-akamai

lgarber-akamai (approved):
Thanks for the contribution!
Generally it is best practice to keep `requirements.txt` and `install_requires` separate, but since we don't pin versions in `requirements.txt` this change should be fine 👍 

Just a note, [a patch for this issue](https://github.com/linode/linode-cli/pull/413) has been merged into the `main` branch and we have a [PR up](https://github.com/linode/linode-cli/pull/415) to merge it back into the `dev` branch. We'll need to merge that change first to keep our git history consistent, so I'll let you know if we need you to resolve any merge conflicts.

chenrui333:
I think this is a better fix of managing deps in two places.

zliang-akamai:
> I think this is a better fix of managing deps in two places.

Yeah, I think that's a good way to manage dependencies. We are going to get the patch in `main` merged back to `dev` then we will be merging this PR.

zliang-akamai (approved):

lgarber-akamai (approved):

chenrui333:
Thanks folks!!

chenrui333:
this approach actually does not quite work as there is requirements.txt in the pypi artifact, either we bundle requirements.txt or we need to revert this change.

chenrui333:
created a patch for such issue (I should really tested before :) )

PR #421, opened 2023-03-24
new: Add support for explicit nullable arguments

### LINKED ISSUE #266: Add ability to clear rDNS
To clear the rDNS for an IP in the web UI, you leave the domain name field blank. This is an illegal operation in the CLI:
```
$ linode-cli networking ip-update
usage: linode-cli [-h] [--rdns rdns] address
linode-cli: error: the following arguments are required: address
$ linode-cli networking ip-update 2600:3c03::1b:1000
Request failed: 400
┌errors─┬──────────────────┐
│ field │ reason           │
├───────┼──────────────────┤
│ rdns  │ rdns is required │
└───────┴──────────────────┘
$ linode-cli networking ip-update --rdns "" 2600:3c03::1b:1000
Request failed: 400
┌errors─┬─────────────────────────────────┐
│ field │ reason                          │
├───────┼─────────────────────────────────┤
│ rdns  │ Length must be 1-255 characters │
└───────┴─────────────────────────────────┘
```

lgarber-akamai:
## 📝 Description

This change allows users to specify `null` as a value for fields that are marked as `nullable` in the API spec. This is useful for situations where the API may have certain functionality tied to having a certain value be explicitly `null` (e.g. RDNS resets)

e.g.
```bash
linode-cli networking ip-update --rdns null 127.0.0.1
```

**NOTE: This will need to be partially reworked after the spec parser refactor, but the change should be relatively straightforward.**

Resolves #266 

## ✔️ How to Test

```
make testunit
```

### DIFF
--- a/linodecli/baked/operation.py
+++ b/linodecli/baked/operation.py
@@ -378,7 +378,7 @@ class OpenAPIOperation:
                         "--" + arg.path,
                         metavar=arg.name,
                         action="append",
-                        type=TYPES[arg.item_type],
+                        type=arg_type_handler,
                     )
                 elif arg.list_item:
                     parser.add_argument(
--- a/linodecli/baked/operation.py
+++ b/linodecli/baked/operation.py
@@ -8,6 +8,7 @@ import platform
 import re
 from getpass import getpass
 from os import environ, path
+from typing import List, Tuple
 
 from linodecli.baked.request import OpenAPIFilteringRequest, OpenAPIRequest
 from linodecli.baked.response import OpenAPIResponse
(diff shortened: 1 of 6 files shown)


linodecli/operation.py
```diff
@@ -6,12 +6,20 @@
...
 import json
 import platform
+from dataclasses import dataclass
 from getpass import getpass
 from os import environ, path
 
 from linodecli.helpers import handle_url_overrides
 
 
+@dataclass
```
zliang-akamai, line 16: `dataclass` decorator seems unnecessary here?

lgarber-akamai: I added `@dataclass` because the linter was complaining about too few public methods with a normal class. It might just be better for me to add a linter ignore here 👍

zliang-akamai (approved):

linodecli/operation.py
zliang-akamai: Haha, I see

jriddle-linode (approved):
Changes look good, will require a couple small changes to spec parser PR but nothing major. Good to merge before parser IMO.

yec-akamai (approved):

zliang-akamai:
Any update on this? @lgarber-akamai

linodecli/baked/operation.py
```diff
@@ -338,6 +363,14 @@ def parse_args(
             for arg in self.args:
                 if arg.read_only:
                     continue
+
+                arg_type_handler = TYPES[
+                    arg.item_type if arg.datatype == "array" else arg.datatype
+                ]
```
zliang-akamai, line 369: I think the `array` is already handled below by the if...else... statement?

lgarber-akamai: So this check is necessary because arrays need to use `arg.item_type` for argparse rather than `arg.datatype`. I moved this out of the `parser.add_argument` call so it could be filtered through the `wrap_parse_nullable_value` function.

zliang-akamai: The array seems to be handled here?
https://github.com/linode/linode-cli/pull/421/files#diff-f69e89596ddac39d63389afca33deb5535e8678cd509684f9d89734007c2f096R375-R382

linodecli/baked/operation.py
```diff
@@ -338,20 +363,29 @@ def parse_args(
...
                     continue
+
+                arg_type = (
+                    arg.item_type if arg.datatype == "array" else arg.datatype
+                )
+                arg_type_handler = TYPES[arg_type]
+
+                if arg.nullable:
+                    arg_type_handler = wrap_parse_nullable_value(arg_type)
+
```
zliang-akamai, line 374: I feel it would be nicer if we can wrap all these logic into an function. But this is just my opinion so it's an optional suggestion!

zliang-akamai (approved):
Everything looks good!

linodecli/baked/operation.py
lgarber-akamai: Good idea!

zliang-akamai (approved):
The refactoring was awesome!

PR #423, opened 2023-03-24
OpenAPI3 Spec Parser

jriddle-linode:
## 📝 Description

Converting the CLI to use the OpenAPI3 Spec Parser this pr will primarily only be changing how cli.bake works and future PRs will add the rest of the features OpenAPI3 offers

## ✔️ How to Test

`make test`

### DIFF
--- a/linodecli/cli.py
+++ b/linodecli/cli.py
@@ -7,6 +7,7 @@ import pickle
 import re
 import sys
 from sys import version_info
+from openapi3 import OpenAPI
 
 from .api_request import do_request
 from .configuration import CLIConfig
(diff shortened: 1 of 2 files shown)


linodecli/api_request.py
```diff
@@ -65,6 +96,37 @@ def do_request(
     return result
 
 
+@staticmethod
```
zliang-akamai, line 99: I think an independent function (not a method of a class) wouldn't need `@staticmethod` decorator.
```suggestion
```

linodecli/api_request.py
```diff
@@ -65,6 +96,37 @@ def do_request(
...
+        merged_result["page"] = 1
+    if "data" in merged_result:
+        merged_result["data"] += list(
+            itertools.chain.from_iterable(r["data"] for r in iterator)
+        )
+    return merged_result
+
+
+def _get_all_pages_results_generator(
+    ctx,
```
zliang-akamai, line 120: Can we add type annotation to `ctx`?

linodecli/arg_helpers.py
```diff
@@ -338,6 +325,18 @@ def action_help(cli, command, action):
             print("You may filter results with:")
             for attr in filterable_attrs:
                 print(f"  --{attr.name}")
+        return
+    if op.args:
```
zliang-akamai, line 329:
```suggestion
    elif op.args:
```

Might be easier to understand this way?

linodecli/baked/colors.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+CLEAR_COLOR = "\x1b[0m"
+COLOR_CODE_MAP = {
+    "red": "\x1b[31m",
+    "green": "\x1b[32m",
+    "yellow": "\x1b[33m",
+    "black": "\x1b[30m",
+    "white": "\x1b[40m",
+}
```
zliang-akamai, line 39: Maybe we can make them all to be constants or enums?

requirements.txt
```diff
@@ -1,5 +1,7 @@
+terminaltables
```
zliang-akamai, line 1: `terminaltables` is an old project and is not under active maintenance. I can't find the usage in the code base. Is there any alternative?

lgarber-akamai: Looks like a remnant from a merge conflict since this branch was created before we migrated from `terminaltables` -> `rich`.

linodecli/baked/colors.py
jriddle-linode: I don't see much of a point in changing this, in-fact we could probably move it to be a variable within the function. Right now its pretty easy to use though.

linodecli/api_request.py
```diff
@@ -65,6 +96,36 @@ def do_request(
...
+    if "page" in merged_result:
+        merged_result["page"] = 1
+    if "data" in merged_result:
+        merged_result["data"] += list(
+            itertools.chain.from_iterable(r["data"] for r in iterator)
+        )
+    return merged_result
+
+
+def _get_all_pages_results_generator(
```
yec-akamai, line 118: nit: the method name sounds a bit odd. Can we align with the style of other methods to name it like `_get_all_pages_results` or `_generate_all_pages_results`? And add a quick description will be helpful as well.

linodecli/api_request.py
```diff
@@ -227,7 +288,7 @@ def _attempt_warn_old_version(ctx, result):
 
         if new_version_exists:
             print(
-                f"The API responded with version {spec_version}, which is newer than "
+                f"!!!!!!!!!!!!!!The API responded with version {spec_version}, which is newer than "
```
yec-akamai, line 291: Lol should we clean it up?

linodecli/baked/request.py
```diff
@@ -0,0 +1,181 @@
...
+    :rtype: list[OpenAPIRequestArg]
+    """
+    args = []
+
+    if schema.properties is not None:
+        for k, v in schema.properties.items():
+            if isinstance(v, Reference):
+                # TODO: can probably remove this at some point,
+                # reference's used to show up in this loop but shouldn't anymore
+                continue
```
yec-akamai, line 103: How to verify if we can remove this or not?

linodecli/baked/response.py
```diff
@@ -0,0 +1,248 @@
...
+
+    @property
+    def path(self):
+        """
+        This is a helper for filterable fields to return the json path to this
+        element in a response.
+        """
+        return self.name
+
+    def _get_value(self, model):
```
yec-akamai, line 88: What is a `model`? Is it a raw response?
Should we use this method to initialize the self.value as well?

yec-akamai (commented):
Great work!

linodecli/baked/response.py
jriddle-linode: added a comment to clarify model is the JSON response. I'm not sure if we can set self.value or not its a remnant from Will's original PR. I've removed it for now.

linodecli/api_request.py
jriddle-linode: Type annotation requires a circular dependency I think. I added comments to clarify though.

linodecli/cli.py
```diff
@@ -42,403 +39,43 @@ def __init__(self, version, base_url, skip_config=False):
...
-                        URLParam(info["name"], info["schema"]["type"])
-                    )
+        ext = {
+            "skip": "linode-cli-skip",
+            "action": "linode-cli-action",
+            "command": "linode-cli-command",
+            "defaults": "linode-cli-allowed-defaults",
+        }
+
+        for _, path in spec.paths.items():
```
zliang-akamai, line 56:
```suggestion
        for path in spec.paths.values():
```

lgarber-akamai (changes requested):
Great work! Just a few fixes and we'll be good to go

lgarber-akamai:
@jriddle-linode I was able to pinpoint the cause of the extension issue with `linode-cli firewalls rules-update` and have put up a PR to address it in `openapi3`: https://github.com/Dorthu/openapi3/pull/102

linodecli/cli.py
```diff
@@ -42,403 +39,43 @@ def __init__(self, version, base_url, skip_config=False):
...
+                self.ops[command][action] = OpenAPIOperation(
+                    command, operation, m, path.parameters
+                )
 
         # hide the base_url from the spec away
-        self.ops["_base_url"] = spec["servers"][0]["url"]
-        self.ops["_spec_version"] = spec["info"]["version"]
+        self.ops["_base_url"] = self.spec.servers[0].url
+        self.ops["_spec_version"] = self.spec.info.version
+        self.ops["_spec"] = self.spec
```
zliang-akamai, line 78: A question confused me... If we are going to drop these when loading the baked data, why do we have to set them here?

zliang-akamai: Ah never mind, I got it

linodecli/cli.py
```diff
@@ -462,6 +99,9 @@ def load_baked(self):
                 if "_spec_version" in self.ops:
                     self.spec_version = self.ops["_spec_version"]
                     del self.ops["_spec_version"]
+                if "_spec" in self.ops:
+                    self.spec = self.ops["_spec"]
+                    del self.ops["_spec"]
```
zliang-akamai, line 104:
```suggestion
                    self.spec = self.ops.pop("_spec")
```
A similar change can also apply to `_base_url` and `_spec_version` above.

zliang-akamai:
Let's fix the tests?

lgarber-akamai (approved):
Good to go! :tada: 

I wonder if we should merge this into a separate upstream branch so we can get unit tests working before release.

zliang-akamai (approved):

PR #424, opened 2023-03-24
add-gha-to-run-integration-tests

ykim-1:
add option to makefile to run all the tests including unit and integration fix test_domain_records.py failures due to target variable purge old bats tests

## 📝 Description

- create new yml file for running integration tests with pytest
- add option to makefile to run all the tests including unit and integration
- fix test_domain_records.py failures due to target variable
- purge old bats tests

## ✔️ How to Test

## 📷 Preview


.github/workflows/e2e-suite.yml
```diff
@@ -16,28 +14,21 @@ jobs:
       - name: Update system packages
         run: sudo apt-get update -y
 
-      - name: Install system deps
-        run: sudo apt-get install -y build-essential bats parallel netcat
-
       - name: Setup Python
         uses: actions/setup-python@75f3110429a8c05be0e1bf360334e4cced2b63fa # pin@v2
```
zliang-akamai, line 18:
```suggestion
        uses: actions/setup-python@v4
```
For official actions published by GitHub, let's pin version number instead of hash

zliang-akamai:
Can you also pin checkout action to be `v3`?

`        uses: actions/checkout@v3` at line 12.

tests/integration/domains/test_domain_records.py
```diff
@@ -38,6 +38,10 @@ def domain_records_setup():
             .rstrip()
         )
 
+    except:
+        logging.exception("Failed creating domain in setup")
```
zliang-akamai, line 42: Having issue with these test cases?

.github/workflows/e2e-suite.yml
```diff
@@ -1,12 +1,10 @@
-name: E2E Test Suite
+name: Integration Tests
 on:
   workflow_dispatch: null
   push:
-    branches:
-      - main
-      - dev
+  pull_request:
```
lgarber-akamai, line 5: For security purposes, this workflow should probably only run on `push` rather than on `push` and `pull_request`. 

To address this, we have a [workflow](https://github.com/linode/linode-cli/blob/dev/.github/workflows/e2e-suite-pr-command.yml) to manually trigger E2E tests when a maintainer uses the `/acctest` command. This is necessary as an external contributor could potentially push up malicious code to grab our test account token, etc. without maintainer approval. Fortunately, GitHub Actions only exposes secrets to workflows that explicitly use them, so our other workflows shouldn't be susceptible to this.

lgarber-akamai, line 5: For security purposes, this workflow should probably only run on `push` rather than on `push` and `pull_request`. 

To address this, we have a [workflow](https://github.com/linode/linode-cli/blob/dev/.github/workflows/e2e-suite-pr-command.yml) to manually trigger E2E tests when a maintainer uses the `/acctest` command. This is necessary as an external contributor could potentially push up malicious code to grab our test account token, etc. without maintainer approval. Fortunately, GitHub Actions only exposes secrets to workflows that explicitly use them, so our other workflows shouldn't be susceptible to this.

lgarber-akamai (changes requested):
Great work! Just one small change

tests/integration/domains/test_domain_records.py
ykim-1: Yeah these were failing - https://linode.slack.com/archives/GPF2AVATC/p1679678683761609

lgarber-akamai:
/acctest sha=a14c7e43adb4d80aef1b9e16b48eb174e988e44c tests=domains

lgarber-akamai (approved):

PR #427, opened 2023-04-03
Add default values for database engines

yec-akamai:
## 📝 Description

Add the default values for database engines in CLI config. https://jira.linode.com/browse/TPT-1676 

## ✔️ How to Test

`make testunit`

Create MySQL and PostgreSQL database separately use default values including engine:
Pasted Graphic

Overwrite the engine value:
image

### DIFF
--- a/linodecli/configuration/__init__.py
+++ b/linodecli/configuration/__init__.py
@@ -336,6 +336,9 @@ If you prefer to supply a Personal Access Token, use `linode-cli configure --tok
         images = [
             i["id"] for i in _do_get_request(self.base_url, "/images")["data"]
         ]
+        engines = [
+            e["id"] for e in _do_get_request(self.base_url, "/engines")["data"]
+        ]
 
         is_full_access = _check_full_access(self.base_url, token)
 
@@ -376,6 +379,13 @@ If you prefer to supply a Personal Access Token, use `linode-cli configure --tok
             images,
             "Default Image (Optional): ",
             "Please select a valid Image, or press Enter to skip",
+        )
+
+        config["engine"] = _default_thing_input(
+            "Default Engine to create a new managed database.",
+            engines,
+            "Default Engine (Optional): ",
+            "Please select a valid Database Engine, or press Enter to skip",
         )
 
         if auth_users:
--- a/linodecli/configuration/__init__.py
+++ b/linodecli/configuration/__init__.py
@@ -212,7 +212,7 @@ class CLIConfig:
     # might be better to move this to argparsing during refactor and just have
     # configuration return defaults or keys or something
     def update(
-        self, namespace, allowed_defaults
+        self, namespace, allowed_defaults, action=None
     ):  # pylint: disable=too-many-branches
         """
         This updates a Namespace (as returned by ArgumentParser) with config values
@@ -242,6 +242,14 @@ class CLIConfig:
                 continue
             if self.config.has_option(username, key):
                 value = self.config.get(username, key)
+            # different types of database creation use different endpoints,
+            # so we need to set the default engine value based on the type
+            elif key == "engine":
+                if action == "mysql-create" and self.config.has_option(username, "mysql_engine"):
+                    value = self.config.get(username, "mysql_engine")
+                elif action == "postgresql-create" and self.config.has_option(username, "postgresql_engine"):
+                    print(self.config.get(username, "postgresql_engine"))
+                    value = self.config.get(username, "postgresql_engine")
             else:
                 value = ns_dict[key]
             if not value:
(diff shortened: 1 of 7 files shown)


linodecli/configuration/helpers.py
```diff
@@ -124,63 +124,69 @@ def _handle_no_default_user(self):
     """
     users = [c for c in self.config.sections() if c != "DEFAULT"]
 
-    if len(users) == 1:
```
yec-akamai, line 127: Refactor code and use case switch to fix a lint issue: `Too many branches`

jriddle-linode (approved):
Looks good just make sure the Docs have been merged and are live before you merge this PR.

linodecli/configuration/helpers.py
zliang-akamai: @yec-akamai `match ... case ...` is the new featured added to Python 3.10. Since we want to maintain backward compatibility to 3.7 or above, can you consider use `if ... else ...` statements here and use another method to get rid of the lint warning, or if it is impossible to avoid, you can disable the lint for this type of warning.

zliang-akamai (changes requested):

linodecli/configuration/helpers.py
yec-akamai: I see. I'll try to figure out another way, but probably have to disable the lint eventually. Thanks for the heads up!

yec-akamai:
> Looks good just make sure the Docs have been merged and are live before you merge this PR.

Created a PR to update the API doc: https://github.com/linode/linode-api-docs/pull/779
Should I reach out to someone to get a review?

linodecli/configuration/__init__.py
```diff
@@ -336,6 +336,9 @@ def configure(
         images = [
             i["id"] for i in _do_get_request(self.base_url, "/images")["data"]
         ]
+        engines = [
+            e["id"] for e in _do_get_request(self.base_url, "/engines")["data"]
```
zliang-akamai, line 340:
```suggestion
            e["id"] for e in _do_get_request(self.base_url, "/databases/engines")["data"]
```
404 error occurs during `linode configure`, I would suggest double check the API endpoint.

zliang-akamai (changes requested):

lgarber-akamai (changes requested):
Hmm, this feature might be a bit trickier than I first thought considering each database type (e.g. MySQL, Postgres) has its own specific endpoint. This can be problematic since the `engine` default would apply to both endpoints, meaning the API would throw an error if a user tried to create a MySQL database using a PostgreSQL engine default (e.g. `postgresql/10.23`). 

We could definitely add some configuration logic to set defaults for each DB type independently using filters, but we'll also need to figure out a way to make some defaults endpoint-specific. Maybe config default keys could have an optional prefix that specifies the endpoint path?

Let me know if you have any ideas!

linodecli/configuration/__init__.py
```diff
@@ -242,6 +242,17 @@ def update(
                 continue
             if self.config.has_option(username, key):
                 value = self.config.get(username, key)
+            # different types of database creation use different endpoints,
+            # so we need to set the default engine value based on the type
+            elif key == "engine":
```
yec-akamai, line 247: `engine` hasn't been added to the doc as `x-linode-cli-allowed-defaults` officially. We have a sparate open PR to do this job.
Before https://github.com/linode/linode-api-docs/pull/779 gets merged and live, we can test the change by hardcoding `engine` in the `allowed_defaults` (L234):
`allowed_defaults.append("engine")`

yec-akamai:
> Hmm, this feature might be a bit trickier than I first thought considering each database type (e.g. MySQL, Postgres) has its own specific endpoint. This can be problematic since the `engine` default would apply to both endpoints, meaning the API would throw an error if a user tried to create a MySQL database using a PostgreSQL engine default (e.g. `postgresql/10.23`).
> 
> We could definitely add some configuration logic to set defaults for each DB type independently using filters, but we'll also need to figure out a way to make some defaults endpoint-specific. Maybe config default keys could have an optional prefix that specifies the endpoint path?
> 
> Let me know if you have any ideas!

Great point! I pass the `aciton` in when building the api request body, so we can tell which database type that user wants to create. I also separate the engine defaults when configuring linode. User can set up default db engine individually for each type of database. It seems that MongoDB is not available for creation currently, right?

linodecli/configuration/__init__.py
lgarber-akamai: > It seems that MongoDB is not available for creation currently, right?

Yep, that's correct 👍

lgarber-akamai:
> It seems that MongoDB is not available for creation currently, right?

Yep, that's correct 👍

linodecli/configuration/__init__.py
```diff
@@ -242,6 +242,14 @@ def update(
...
+                if action == "mysql-create" and self.config.has_option(username, "mysql_engine"):
```
lgarber-akamai: This looks really good! 
Maybe in the future we could optionally accept `x-linode-cli-allowed-defaults` as a dict so config default mappings could be handled in the spec.

e.g.
```yaml
x-linode-cli-allowed-defaults:
    engine: mysql_engine
```

lgarber-akamai (approved):
Tested locally and everything works as intended. Nice work!

linodecli/configuration/__init__.py
yec-akamai: Oh thanks for sharing this trick! It's much better than hardcoding the value lol

linodecli/configuration/__init__.py
yec-akamai: Oh good idea. I didn't realize it can work this way. Let me create a follow-up ticket to optimize in the future.

zliang-akamai (approved):

PR #429, opened 2023-04-03
Move objects and buckets methods to separate files

zliang-akamai:
## 📝 Description

Make them more readable.

## ✔️ How to Test

`pytest tests/integration/test_obj_plugin.py`

### DIFF
--- a/linodecli/plugins/obj/buckets.py
+++ b/linodecli/plugins/obj/buckets.py
@@ -1,6 +1,5 @@
-# pylint: disable=too-many-lines
 """
-CLI Plugin for handling OBJ
+The bucket manipulation module of CLI Plugin for handling object storage
 """
 import sys
 from argparse import ArgumentParser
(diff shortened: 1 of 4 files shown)


jriddle-linode (approved):

linodecli/plugins/obj/buckets.py
```diff
@@ -0,0 +1,78 @@
+# pylint: disable=too-many-lines
+"""
+CLI Plugin for handling OBJ
```
yec-akamai, line 3: nit: should the comment be CLI Plugin for handling buckets, or obj buckets?

yec-akamai (commented):
Should we also consider refactoring the test file to separate objects and buckets? It's probably easier to find the corresponding test cases in the future.

linodecli/plugins/obj/buckets.py
zliang-akamai: Good catch! Thank you for the suggestion, and I just updated the doc string accordingly.

lgarber-akamai:
/acctest sha=53e33359f0f0227c469d3309a08ff1d2c37ef59e tests=test_obj_plugin.py

lgarber-akamai (approved):
Nice work!

PR #433, opened 2023-04-04
Fix/teardown pytest

ykim-1:
## 📝 Description

Fixing teardowns in cli test suites

## ✔️ How to Test

make testint 

## 📷 Preview

### DIFF
--- a/tests/integration/domains/test_domain_records.py
+++ b/tests/integration/domains/test_domain_records.py
@@ -87,21 +87,25 @@ def test_create_a_domain():
     output_current = process.stdout.decode()
 
     # Create domain
-    domain_id = exec_test_command(
-        BASE_CMD
-        + [
-            "create",
-            "--type",
-            "master",
-            "--domain",
-            timestamp + "example.com",
-            "--soa_email=pthiel@linode.com",
-            "--text",
-            "--no-header",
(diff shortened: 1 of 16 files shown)


jriddle-linode (approved):

tests/integration/domains/test_domain_records.py
```diff
@@ -112,6 +118,8 @@ def test_create_a_domain():
         "the list is not updated with new domain..",
     )
 
+    delete_target_id(target="domains", id=domain_id)
```
zliang-akamai, line 121: A concern I have is that for some test cases like this one, if the previous assertion failed before reaching the cleanup part of the code, the created resources may remain on the cloud and required a manual cleanup... I think an ideal way is to put creation command invocation in a fixture, similar to this example below. Or can you think about another way for this kind of situation?
https://github.com/linode/linode-cli/blob/0e5208c8b3ce89eb3da020a320124bc4df709b72/tests/integration/test_obj_plugin.py#L90

ykim-1: Thanks for this catch. You are right teardowns/cleanups should not happen after assertion

zliang-akamai (approved):
Good work!

PR #438, opened 2023-04-17
migrate unittest to pytest

ykim-1:
## 📝 Description

removed unittest library dependencies in all unit tests
removed unittest command from Makefile

## ✔️ How to Test

make testunit

## 📷 Preview


jriddle-linode (approved):

.github/workflows/unit-tests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           python-version: '3.x'
 
       - name: Install Python wheel
-        run: pip install wheel boto3
+        run: pip install wheel boto3 mock
```
lgarber-akamai, line 23: We might be able to move this dependency to `requirements-dev.txt`

lgarber-akamai (changes requested):
Looks great! Just a few small things.

NOTE: It looks like `pytest` isn't picking up test files that don't contain the `test_` prefix. Could files without the prefix have one added?

tests/unit/configuration.py
```diff
@@ -103,10 +109,13 @@ def test_print_users(self):
...
-            conf.print_users()
-        self.assertEqual(cm.exception.code, 0)
-        self.assertTrue("*  cli-dev" in f.getvalue())
+
+        try:
+            with contextlib.redirect_stdout(f):
+                conf.print_users()
+        except SystemExit as err:
+            assert err.code == 0
+            assert "*  cli-dev" in f.getvalue()
```
zliang-akamai, line 118: https://docs.pytest.org/en/latest/how-to/assert.html#assertions-about-expected-exceptions

Might be better to use `pytest.raises` context manager here.

ykim-1: Think this should be ok for now since it is actually catching the error and asserting properly

tests/unit/conftest.py
```diff
@@ -41,7 +40,7 @@ def mock_cli(
     # We need this to suppress warnings for operations that don't
     # have access to the cli.suppress_warnings attribute.
     # e.g. operation defaults
-    sys.argv.append("--suppress-warnings")
+    # sys.argv.append("--suppress-warnings")
```
yec-akamai, line 43: Maybe we can add a bit context on why it needs to be removed here for visibility.

tests/unit/test_completion.py
```diff
@@ -48,37 +47,38 @@ class CompletionTests(unittest.TestCase):
...
         actual = completion.get_fish_completions(self.ops)
-        self.assertEqual(actual, self.fish_expected)
+        assert actual == self.fish_expected
 
-    def test_bash_completion(self):
+    def test_bash_completion(self, mocker):
         """
         Test if the bash completion renders correctly
         """
+        # mocker = mocker.patch('linodecli-completion.get_bash_completions', return_value=self.bash_expected)
```
yec-akamai, line 61: nit: it seems can be cleaned up.

yec-akamai (approved):
LGTM!

ykim-1:
> Looks great! Just a few small things 🎉
> 
> NOTE: It looks like `pytest` isn't picking up test files that don't contain the `test_` prefix. Could files without the prefix have one added?

Changed the file naming instead for consistency. E.g. configuration.py -> test_configuration.py

PR #440, opened 2023-04-18
Replace terminaltables with rich

noprobelm:
## 📝 Description

Relates to #436 

The version of `terminaltables` used by this project has been archived for a few years now and was recently removed from the Arch community repository (which is what led me to embark on this PR). While the project may remain viable for some time to come, I propose the maintainers of `linode-cli` consider switching to [rich](https://github.com/textualize/rich), which offers much greater versatility when rendering rich text to the terminal in Python (even `pip` uses rich for its progress bar functionality).

I have made the necessary changes to the code in this PR.

### Summary of changes:
- Updated `requirements.txt`
  - Added `rich`
  - Removed `terminaltables`
- Updated all instances of `terminaltables.SingleTable` objects throughout the project with equivalent `rich.Table` objects. The following modules were updated under this effort:
  - `arg_helpers.py`
  - `output.py` - Note for this module I also added the `rich.Text` module to enable proper ANSI sequence rendering. Without converting the table cells to `Text` objects, any cell containing ANSI sequences will result in improper rendering of the columns. This should not be applicable to any other module, but could be implemented for uniformity's sake if desired with no impact to functionality.
  - `__init__.py`
  - `plugins/obj/__init__.py`
  - `plugins/obj/helpers.py`
- Updated the `firewall-editor` plugin to replace `terminaltables.PorcelainTable` object with `rich.Table`. The new `rich.Table` is configured to be equivalent to the corresponding `PorcelainTable` it's replacing.
- Updated the unit test `test_output.py` to use `rich.Table`

### Steps
- Run any set of commands associated with generating tables. Among these, make sure to run a command that generates a table with colored text, such as `linode-cli linodes list` to ensure the ANSI escape sequence'd text renders correctly.
- Run `linode-cli` to see the main help message tables
- Run `linode-cli firewall-editor <label>` to see the new firewall editor table.
- Run the `test_output.py` unit test `python -m unittest test_output.py`

### What I tested
- Each of the 4 tests above

### What I **did not** test
- I did  not run the full test suite as I didn't want to destroy my account's configuration nor did I want to make a new account for this purpose. 

### Before/After
- `linode-cli` before
linode-cli-help-before
- `linode-cli` after
linode-cli-help-after
- `linode-cli linodes list` before
linode-cli-linodes-list-before
- `linode-cli linodes list` after
linode-cli-linodes-list-after
- `linode-cli firewall-editor <label>` before
linode-cli-firewall-editor-before
- `linode-cli firewall-editor <label>` after
[...]

PR #443, opened 2023-04-20
bug: Shorten datetime strings output from `linode-cli obj ls` operation

Dorthu:
Noticed this while reviewing #440.

I suspect that `boto` used to output dates like this, and the format
changed in the conversion to `boto3`.

Current output:

```
$ linode-cli obj ls
2021-03-10 12:32:10.317000+00:00  example-bucket
```

This branch:

```
$ linode-cli obj ls
2021-03-10 12:32  example-bucket
```

`s3cmd ls` output (for reference - this plugin was originally based on
that):

```
$ s3cmd ls
2021-03-10 12:32  s3://example-bucket
```

### DIFF
--- a/linodecli/plugins/obj/__init__.py
+++ b/linodecli/plugins/obj/__init__.py
@@ -121,7 +121,7 @@ def list_objects_or_buckets(
             if key == prefix:
                 continue
 
-            data.append((obj.get("LastModified"), obj.get("Size"), key))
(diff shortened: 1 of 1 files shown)


lgarber-akamai (approved):
Seems to have resolved the issue on my end 🎉

zliang-akamai (approved):

PR #449, opened 2023-05-01
new: Add support for output overrides and add `domains zone-file` override

lgarber-akamai:
## 📝 Description

This change introduces a new output override system that allows us to modularly add custom output logic for specific generated commands. This is useful for cleaning up outputs for endpoints with responses that may be too complicated to account for in the spec parser.

### DIFF
--- a/linodecli/operation.py
+++ b/linodecli/operation.py
@@ -9,7 +9,8 @@ import platform
 from getpass import getpass
 from os import environ, path
 
-from linodecli import output_overrides
+from .overrides import OUTPUT_OVERRIDES
+
 from linodecli.helpers import handle_url_overrides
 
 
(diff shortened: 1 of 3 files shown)


linodecli/operation.py
```diff
@@ -346,6 +348,12 @@ def process_response_json(
         if self.response_model is None:
             return
 
+        override = OUTPUT_OVERRIDES.get(
+            (self.command, self.action, handler.mode)
+        )
+        if override is not None and not override(self, handler, json):
+            return
```
zliang-akamai, line 355: Wonder why we need to check `not override(self, handler, json)` here. For example, if the override function return a `True` then we will have to run the rest of `process_response_json` function?

lgarber-akamai: > For example, if the override function return a True then we will have to run the rest of process_response_json function?

Yep! We're checking for the result to be `False` here so we can exit the function early if necessary

zliang-akamai (approved):

ezilber-akamai (approved):

PR #450, opened 2023-05-01
Fix error assertions and cleanup dependencies

ykim-1:
## 📝 Description

replaced all try and except blocks with pytest.raises for cleaner and proper assertion
since mock is part of python std lib, we can import it directly from unittest

## ✔️ How to Test

## 📷 Preview

### DIFF
--- a/tests/unit/test_configuration.py
+++ b/tests/unit/test_configuration.py
@@ -73,10 +73,11 @@ mysql_engine = mysql/8.0.26"""
 
         f = io.StringIO()
 
-        with pytest.raises(SystemExit):
+        with pytest.raises(SystemExit) as err:
             with contextlib.redirect_stdout(f):
                 conf.set_user("bad_user")
 
+        assert err.value.code == 1
         assert "not configured" in f.getvalue()
 
         conf.set_user("cli-dev2")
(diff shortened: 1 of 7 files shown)


tests/unit/test_configuration.py
```diff
@@ -110,12 +108,11 @@ def test_print_users(self):
 
         f = io.StringIO()
 
-        try:
+        with pytest.raises(SystemExit):
             with contextlib.redirect_stdout(f):
                 conf.print_users()
```
zliang-akamai, line 113:
```suggestion
        with pytest.raises(SystemExit) as excinfo:
            with contextlib.redirect_stdout(f):
                conf.print_users()
        assert excinfo.value.code == 0
```

Can you try something like this to assert the error code of the `SystemExit` instance?

The pytest API can allow you to access the instance of the exception, you can see [details docs here](https://docs.pytest.org/en/latest/how-to/assert.html#assertions-about-expected-exceptions).

ykim-1: Done. Don't know why I added error code assertion in other places but not this file lol..

zliang-akamai (approved):

zliang-akamai:
Nice work!

yec-akamai (approved):
lgtm

PR #455, opened 2023-05-09
Add `--all-rows` Flag

zliang-akamai:
## 📝 Description

Provide a option for user to output all results without pagination.

Should we add some integration/unit tests for this? Will be working on that if necessary.

## ✔️ How to Test

Create 1000 stackscripts in your account then run `linode --all-rows stackscripts ls --is_public false`

### DIFF
--- a/linodecli/__init__.py
+++ b/linodecli/__init__.py
@@ -77,8 +77,16 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
         cli.output_handler.columns = "*"
     if parsed.no_headers:
         cli.output_handler.headers = False
-    if parsed.all:
+    if parsed.all_columns or parsed.all:
+        if parsed.all:
+            print(
+                "WARNING: '--all' is a deprecated flag, "
+                "and will be removed in a future version. "
+                "Please consider use '--all-columns' instead."
+            )
         cli.output_handler.columns = "*"
+    if parsed.all_rows:
+        cli.pagination = False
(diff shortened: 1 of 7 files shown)


linodecli/cli.py
```diff
@@ -479,6 +482,59 @@ def _flatten_url_path(tag):
...
+
+            result = self.merge_results_data(
+                itertools.chain(
+                    (result,),
+                    self._get_all_pages_results_generator(
+                        operation, args, pages_needed
+                    ),
+                )
+            )
+        return result
```
lgarber-akamai, line 536: This looks great! Could we add some unit tests to make sure all of these function work as expected with some mock data?

zliang-akamai: Thank you! Test added.

linodecli/__init__.py
```diff
@@ -77,8 +77,16 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
         cli.output_handler.columns = "*"
     if parsed.no_headers:
         cli.output_handler.headers = False
-    if parsed.all:
+    if parsed.all_columns or parsed.all:
+        if parsed.all:
```
lgarber-akamai, line 81: Could we check whether warnings should be suppressed here?

lgarber-akamai (changes requested):
Works as expected on my end. One small change and we should be good to go 👍

lgarber-akamai (approved):

PR #456, opened 2023-05-10
new: Improve handling for list structures

lgarber-akamai:
## 📝 Description

This change improves the handling for nested list structures (e.g. Instance Interfaces) by implicitly detecting when the next list entry should be populated. This means that users do not have to explicitly specify empty values for null fields.

For example, see this command:

```bash
linode-cli linodes config-update 1234 5678 \
--interfaces.purpose public \
--interfaces.purpose vlan --interfaces.label my-vlan 
```

A user would intuitively expect this command to set the first interface to public and the second interface to a VLAN with the label `my-vlan`, however executing this command would generate this request body:

```json
{
   "interfaces":[
      {
         "label":"cool",
         "purpose":"public"
      }
   ]
}
```

After this change, the request body is now generated as expected:

```json
{
   "interfaces":[
      {
         "label":null,
         "ipam_address":null,
         "purpose":"public"
      },
      {
         "label":"cool",
         "purpose":"vlan"
      }
   ]
}
```

## ✔️ How to Test

```
make test
```

### DIFF
--- a/linodecli/operation.py
+++ b/linodecli/operation.py
@@ -113,12 +113,19 @@ class ListArgumentAction(argparse.Action):
 
         dest_list = getattr(namespace, self.dest)
         dest_length = len(dest_list)
+        dest_parent = self.dest.split(".")[:-1]
+
+        # If this isn't a nested structure,
+        # append and return early
+        if len(dest_parent) < 1:
+            dest_list.append(values)
+            return
 
         # A list of adjacent fields
         adjacent_keys = [
             k
             for k, v in namespace.__dict__.items()
-            if k.split(".")[:-1] == self.dest.split(".")[:-1]
+            if k.split(".")[:-1] == dest_parent
         ]
 
         # Let's populate adjacent fields ahead of time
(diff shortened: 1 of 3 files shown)


yec-akamai (approved):
Nice work!

linodecli/operation.py
```diff
@@ -100,6 +100,58 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+
+        # If we're creating a new list object, append
+        # None to every non-populated field.
+        if dest_length >= deepest_length:
+            for k, item in adjacent_items.items():
+                if k == self.dest:
+                    continue
+
+                if len(item) < dest_length:
+                    item.append(None)
```
zliang-akamai, line 150: I think you meant while loop here?
```suggestion
                while len(item) < dest_length:
                    item.append(None)
```

linodecli/operation.py
```diff
@@ -100,6 +100,58 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+        # If this isn't a nested structure,
+        # append and return early
+        if len(dest_parent) < 1:
+            dest_list.append(values)
+            return
+
+        # A list of adjacent fields
+        adjacent_keys = [
+            k
+            for k, v in namespace.__dict__.items()
```
zliang-akamai, line 127:
```suggestion
            for k, v in vars(namespace).items()
```

zliang-akamai (approved):

PR #462, opened 2023-05-31
update readme file to reflect changes in integration tests

ykim-1:
## 📝 Description

update readme to reflect migration of integration tests from bats to pytest

Also gave some options in Makefile to allow users to run specific test suite/case/package

## ✔️ How to Test

## 📷 Preview


zliang-akamai:
How can we run the tests directly in the `tests/integration` directory, such as `tests/integration/test_obj_plugin.py`. Can we place them in a module to solve it?

Makefile
```diff
@@ -1,8 +1,22 @@
 #
 # Makefile for more convenient building of the Linode CLI and its baked content
 #
+PKG_PATH =
```
zliang-akamai, line 4: I think it might be better to name it as `MOD_PATH` as well as other vars, because in Python, a sub-directory is called a module, not a package like what it's called in Golang.

Makefile
```diff
@@ -1,8 +1,22 @@
...
+TEST_SUITE_PATH =
+TEST_CASE_COMMAND =
```
zliang-akamai, line 6: Can we have some explicit default values for these vars?

zliang-akamai: Maybe some empty strings, `""`?

zliang-akamai:
And if we are updating the Makfile, I think we also need to update the acctest command workflow
https://github.com/linode/linode-cli/blob/dev/.github/workflows/e2e-suite-pr.yml#L50-L54

Makefile
ykim-1: It is best left null for these variables because `testint` command will actually run the entire test suite when left empty. However if we give some empty string value such as "", pytest will not function properly.
For reference, 	`pytest tests/integration$(PKG_PATH)$(TEST_SUITE_PATH) $(TEST_CASE_COMMAND)
`

ykim-1:
> How can we run the tests directly in the `tests/integration` directory, such as `tests/integration/test_obj_plugin.py`. Can we place them in a module (sub-directory) to solve it?

Yeah, I put them in plugin directory

yec-akamai (approved):

PR #472, opened 2023-06-13
new: Add a `No Default` option when running the `configure` command

lgarber-akamai:
## 📝 Description

This change adds a `No Default` option to the `linode-cli configure` command that allows users to optionally clear a default if it is already defined in the existing config. This differs from the skip (enter) functionality in that the skip functionality retains the previous config value and the `No Default` option drops the config value.

## ✔️ How to Test

`make testunit`

### DIFF
--- a/linodecli/configuration/__init__.py
+++ b/linodecli/configuration/__init__.py
@@ -424,7 +424,7 @@ If you prefer to supply a Personal Access Token, use `linode-cli configure --tok
             "Default Engine (Optional): ",
             "Please select a valid PostgreSQL Database Engine, or press Enter to skip",
             current_value=_config_get_with_default(
-                self.config, username, "postgres_engine"
+                self.config, username, "postgresql_engine"
             ),
         )
 
--- a/linodecli/configuration/__init__.py
+++ b/linodecli/configuration/__init__.py
@@ -15,6 +15,7 @@ from .auth import (
 )
 from .helpers import (
     _check_browsers,
+    _config_get_with_default,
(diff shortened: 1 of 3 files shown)


ezilber-akamai (approved):
LGTM!

yec-akamai (commented):
Code looks good! However, when I was testing locally, there isn't a `No default` option popping up for `postgresql`, even though i set its default value in a previous configure. Others work well. I can't tell why this value is different. Do you have any idea on it?

lgarber-akamai:
> Code looks good! However, when I was testing locally, there isn't a `No default` option popping up for `postgresql`, even though i set its default value in a previous configure. Others work well. I can't tell why this value is different. Do you have any idea on it?

Let me take a look :eyes:

lgarber-akamai:
> Code looks good! However, when I was testing locally, there isn't a `No default` option popping up for `postgresql`, even though i set its default value in a previous configure. Others work well. I can't tell why this value is different. Do you have any idea on it?

Ah okay, looks like it was just a typo. Could you test this again with the latest commit @yec-akamai ?

yec-akamai:
> > Code looks good! However, when I was testing locally, there isn't a `No default` option popping up for `postgresql`, even though i set its default value in a previous configure. Others work well. I can't tell why this value is different. Do you have any idea on it?
> 
> Ah okay, looks like it was just a typo. Could you test this again with the latest commit @yec-akamai ?

Works now! Thank you

yec-akamai (approved):

PR #476, opened 2023-06-20
fix: Use `+and` for X-Filter header and support filtering on multiple list entries

lgarber-akamai:
## 📝 Description

This change adds support for filtering on multiple entries of the same response list when calling list endpoints. This works by flattening out any parsed list arguments and appending all values to a `+and` filter.

For example:

```bash
linode-cli linodes ls --tags foo --tags bar
```

This command would previously ignore the first specified tag and only filter on the second tag. With this fix, response rows will be required to have both specified tags.

## ✔️ How to Test

```bash
make testunit
```

### DIFF
--- a/linodecli/api_request.py
+++ b/linodecli/api_request.py
@@ -77,18 +77,30 @@ def _build_filter_header(
     if filter_header is not None:
         return json.dumps(filter_header)
 
-    filters = vars(parsed_args)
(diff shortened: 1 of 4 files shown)


zliang-akamai (approved):

ykim-1 (approved):
Good work, looks good!

PR #477, opened 2023-06-21
Migrate build frontend to `pypa/build`

zliang-akamai:
## 📝 Description

Calling `setup.py` directly is now deprecated.


lgarber-akamai (approved):
Works locally after manually re-running `pip install -r requirements-dev.txt`. Nice work!

yec-akamai (approved):

PR #493, opened 2023-07-14
new: Allow API target overrides through the config file

lgarber-akamai:
## 📝 Description

This change allows for the API target to be overridden through entries in the config file. These fields can be configured during the `linode-cli` configure process.

Additionally, this change introduces helpers for boolean and text inputs during the configuration process.

This PR also fixes a few bugs and broken tests I discovered in the `openspec` branch. This includes a bug that would break request URLs when a custom `api_version` is specified.

## ✔️ How to Test

```
make testunit
```


jriddle-linode (approved):

yec-akamai (approved):
Nice work!

PR #504, opened 2023-08-02
All rows flag for obj plugin

zliang-akamai:
## 📝 Description

With it the user may get all objects in a bucket now.
I also plan to wrap up all list functionalities in a separated file in the next PR.

## ✔️ How to Test

You can add more than 1000 files into a bucket and the test the flag.

### DIFF
--- a/linodecli/plugins/obj/__init__.py
+++ b/linodecli/plugins/obj/__init__.py
@@ -22,7 +22,7 @@ from rich.table import Table
 from linodecli.cli import CLI
 from linodecli.configuration import _do_get_request
 from linodecli.configuration.helpers import _default_thing_input
-from linodecli.helpers import expand_globs, pagination_args
+from linodecli.helpers import expand_globs, pagination_args_shared
 from linodecli.plugins import PluginContext, inherit_plugin_args
 from linodecli.plugins.obj.buckets import create_bucket, delete_bucket
 from linodecli.plugins.obj.config import (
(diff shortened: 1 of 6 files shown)


linodecli/plugins/obj/__init__.py
```diff
@@ -106,16 +132,28 @@ def list_objects_or_buckets(
...
+            if parsed.all_rows:
+                responses = client.get_paginator("list_objects_v2").paginate(
+                    Prefix=prefix, Bucket=bucket_name, Delimiter="/"
+                )
+            else:
+                response = client.list_objects_v2(
+                    Prefix=prefix, Bucket=bucket_name, Delimiter="/"
+                )
+                if response.get("IsTruncated", False):
+                    warn(parsed, TRUNCATED_WARNING_MSG)
```
lgarber-akamai, line 147: I wonder if it'd be useful for users to be able to configure the page size for these requests. This might also come in handy for testing since we wouldn't need to create a bucket with 1000+ objects to test pagination.

Just a thought, feel free to ignore :slightly_smiling_face:

zliang-akamai: Haha, that make sense. Thank you for the feedback! For now you can try to pass in a `PageSize` parameter for testing. I will add it to the PR later.

yec-akamai (commented):
LGTM. Can we add unit test as well?

lgarber-akamai (commented):
Is there any chance we could also add an integration test for this?
Maybe something like:
- Create a bucket
- Populate it with 5 objects
- Set the page size to 2
- Ensure all 5 objects are returned with the `--allow-rows` flag

amisiorek-akamai (approved):
Looks good!

PR #505, opened 2023-08-04
new: retry on common errors

jriddle-linode:
## 📝 Description

Adding retry functionality for common random errors sent by the API

- 400: Bad Request with nginx headers
- 408: Request Timeout
- 429: Rate limit exceeded

## ✔️ How to Test

`make test`

### DIFF
--- a/linodecli/api_request.py
+++ b/linodecli/api_request.py
@@ -4,6 +4,7 @@ This module is responsible for handling HTTP requests to the Linode API.
 
 import itertools
 import json
+import time
 import sys
 from sys import version_info
 from typing import Iterable, List, Optional
@@ -12,9 +13,9 @@ import requests
 from packaging import version
 from requests import Response
 
+from linodecli.helpers import API_CA_PATH
 from .baked.operation import OpenAPIOperation
 from .helpers import handle_url_overrides
-from linodecli.helpers import API_CA_PATH
 
 
 def get_all_pages(ctx, operation: OpenAPIOperation, args: List[str]):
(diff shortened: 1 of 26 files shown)


ykim-1 (approved):
Looks good!

zliang-akamai:
In Python SDK we used [the retry class from urllib3](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry). Should we consider that for linode-cli as well or?

linodecli/api_request.py
```diff
@@ -339,3 +345,33 @@ def _handle_error(ctx, response):
...
+    return False
+
+
+def _get_retry_after(headers):
+    if headers.get("Retry-After"):
+        retry_str = headers.get("Retry-After")
+        if retry_str == "":
+            return 0
+        return int(retry_str)
+    return 0
```
amisiorek-akamai, line 377: Looks good! Just a small suggestion.

```suggestion
    retry_str = headers.get("Retry-After", "")
    return int(retry_str) if retry_str else 0
```

lgarber-akamai:
> In Python SDK we used [the retry class from urllib3](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry). Should we consider that for linode-cli as well or?
> 
> https://github.com/linode/linode_api4-python/blob/dev/linode_api4/linode_client.py#L26

That's a great question! The biggest limitation I can think of with that approach is that it might be a good bit more complicated to implement complex retry logic (e.g. 400 NGINX errors)

linodecli/api_request.py
```diff
@@ -339,3 +345,33 @@ def _handle_error(ctx, response):
...
+        return True
+    if response.headers:
+        if (
+            response.status_code == 400
+            and response.headers.get("Server") == "nginx"
+            and response.headers.get("Content-Type") == "text/html"
+        ):
+            # nginx html response
+            return True
+    return False
```
lgarber-akamai, line 368: Just a tiny optional code suggestion

```suggestion
    if response.status_code in (408, 429):
        # request timed out or rate limit exceeded
        return True

    if (
        response.headers 
        and response.status_code == 400
        and response.headers.get("Server") == "nginx"
        and response.headers.get("Content-Type") == "text/html"
    ):
        # nginx html response
        return True

    return False
```

lgarber-akamai (approved):
Looks good to me! Just needs a `make format` run .

PR #506, opened 2023-08-09
Add order by and order flag

amisiorek-akamai:
## 📝 Description

Adds --order_by and --order to filtering operations.

## ✔️ How to Test

Tested using command line. 

Ex.
`linode-cli linodes list --order_by label --order desc`
`linode-cli linodes list --order_by label`

### DIFF
--- a/linodecli/api_request.py
+++ b/linodecli/api_request.py
@@ -151,16 +151,13 @@ def _build_filter_header(
             del parsed_args_dict[p.name]
 
     # check for order_by and order
-    order_by = None
+    order_by, order = None, None
     if parsed_args_dict['order_by'] is not None:
         order_by = parsed_args_dict['order_by']
         order = 'asc' if parsed_args_dict['order'] == None else parsed_args_dict['order']
     
     del parsed_args_dict['order_by']
     del parsed_args_dict['order']
-
-    print(order_by, order)
-
 
     # The "+and" list to be used in the filter header
     filter_list = []
--- a/linodecli/api_request.py
+++ b/linodecli/api_request.py
@@ -150,6 +150,18 @@ def _build_filter_header(
         if p.name in parsed_args_dict:
             del parsed_args_dict[p.name]
 
+    # check for order_by and order
+    order_by = None
+    if parsed_args_dict['order_by'] is not None:
+        order_by = parsed_args_dict['order_by']
+        order = 'asc' if parsed_args_dict['order'] == None else parsed_args_dict['order']
+    
+    del parsed_args_dict['order_by']
+    del parsed_args_dict['order']
+
+    print(order_by, order)
+
+
     # The "+and" list to be used in the filter header
     filter_list = []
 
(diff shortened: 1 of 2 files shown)


ykim-1:
Looks like unit tests and lint are failing. Tests might need to be updated

linodecli/api_request.py
```diff
@@ -150,6 +150,19 @@ def _build_filter_header(
...
+        order_by = parsed_args_dict["order_by"]
+        order = (
+            "asc"
+            if parsed_args_dict["order"] is None
+            else parsed_args_dict["order"]
+        )
+
+    del parsed_args_dict["order_by"]
+    del parsed_args_dict["order"]
+
```
lgarber-akamai, line 165: This logic can be simplified down a bit:

```suggestion
    # check for order_by and order
    order_by = parsed_args_dict.pop("order_by")
    order = parsed_args_dict.pop("order") or "asc"
```
See below for the corresponding suggestion.

linodecli/api_request.py
```diff
@@ -162,7 +175,13 @@ def _build_filter_header(
...
-        return json.dumps({"+and": filter_list})
+        if order_by is None:
+            return json.dumps({"+and": filter_list})
+        return json.dumps(
+            {"+and": filter_list, "+order_by": order_by, "+order": order}
+        )
+    if order_by is not None:
+        return json.dumps({"+order_by": order_by, "+order": order})
 
     return None
```
lgarber-akamai, line 186:
```suggestion
    result = {}

    if len(filter_list) > 0:
        result["+and"] = filter_list

    if order_by is not None:
        result["+order_by"] = order_by
        result["+order"] = order

    return json.dumps(result) if len(result) > 0 else None
```

tests/unit/test_api_request.py
```diff
@@ -109,6 +109,8 @@ def test_build_filter_header(self, list_operation):
             SimpleNamespace(
                 filterable_result="bar",
                 filterable_list_result=["foo", "bar"],
+                order_by=None,
+                order=None,
```
lgarber-akamai, line 113: Could we add some extra tests to check the generated header with values specified for order and order_by?

lgarber-akamai (changes requested):
Great work, this seems to work well in my local testing! Just a few small code suggestions and a few missing tests and we should be good to go!

linodecli/api_request.py
amisiorek-akamai: Will definitely do that, that is so much cleaner!

linodecli/api_request.py
amisiorek-akamai: Will likewise change this one too, thanks for the suggestions!

lgarber-akamai (approved):
LGTM!

zliang-akamai (approved):

zliang-akamai:
there seem to be a large conflict to solve...

lgarber-akamai:
Looks like the conflicts are related to the refactors I made to the `parse_args` function: https://github.com/linode/linode-cli/commit/ba57de80f9c8ed66c3935d6f3904c0d275dac63d

amisiorek-akamai:
Just gonna quickly check that the functionality still remains, I see that we just split up parse_args to be a bit neater so shouldn't impact either PR

zliang-akamai (approved):

PR #507, opened 2023-08-11
v5.42.0

zliang-akamai:
~~Should we make it v6 because of `openapi3` being used?~~


lgarber-akamai:
@zliang-akamai I think we can keep this at v5 so we can reserve major version bumps for breaking changes, although that's definitely up for debate

zliang-akamai:
```
---------------------------- Captured stderr setup -----------------------------
The API responded with version 4.160.2, which is newer than the CLI's version of 4.159.0.  Please update the CLI to get access to the newest features.  You can update with a simple `pip3 install --upgrade linode-cli`
----------------------------- Captured stderr call -----------------------------
The API responded with version 4.160.2, which is newer than the CLI's version of 4.159.0.  Please update the CLI to get access to the newest features.  You can update with a simple `pip3 install --upgrade linode-cli`
Request failed: 400
X-Filter,Cannot filter on entity.id
X-Filter,Cannot filter on entity.type
```
This is a known issue, I think.

jriddle-linode (approved):
Exciting.

PR #511, opened 2023-08-16
proposal: Add support for printing additional tables in command outputs

### LINKED ISSUE #228: linodes list-ips output not working as expected for --json and --text
Using linode-cli v5.0.1, when I run the command:

linode-cli linodes ips-list --json --suppress-warnings --format 'private' 123456

I get some json-like outpout.  But when I do it with '--text --no-headers' I get shorter more targeted output but also in json format.  

For example, --json first.

```
$ linode-cli linodes ips-list --suppress-warnings --format 'private' --json 25152787
[{"ipv4": {"private": [{"address": "192.168.215.203", "gateway": null, "subnet_mask": "255.255.128.0", "prefix": 17, "type": "ipv4", "public": false, "rdns": null, "linode_id": 25152787, "region": "us-central"}]}}]
```

Now with '--text --no-headers' instead:

```
$ linode-cli linodes ips-list --suppress-warnings --format 'private' --text --no-headers 25152787
{'address': '192.168.215.203', 'gateway': None, 'subnet_mask': '255.255.128.0', 'prefix': 17, 'type': 'ipv4', 'public': False, 'rdns': None, 'linode_id': 25152787, 'region': 'us-central'}
```

See how it comes out in JSON mode?  This is actually what I want the first format to be, since it's simpler to parse within Ansible.

lgarber-akamai:
## 📝 Description

This change adds support for dynamically printing subtables in outputs for API endpoints with complex nested structures. These subtables are indicated through a new `x-linode-cli-subtables` OpenAPI spec extension.

This PR will need a follow-up change to the [Linode OpenAPI spec](https://github.com/linode/linode-api-docs). For the time being, I've embedded a modified `openapi.yaml` in this PR.

Additionally, this PR adds two new CLI arguments:
- `--single-table` - Disable printing multiple tables for complex API responses.
- `--table` - The specific table(s) to print in output of a command.

Known issues:
- The `linode-cli linodes ips-list {linodeId}` output is not parsable in single-table mode. This issue was caused by the spec parser migration and was not the result of this change. This might not be worth addressing since automation should generally use `--json` (in conjunction with `jq` in most cases). 

Blocked by https://github.com/Dorthu/openapi3/pull/109

Resolves #228 

## ✔️ How to Test

```
make SPEC=$PWD/openapi.yaml install
linode-cli linodes ips-list {instanceId}
linode-cli firewalls rules-list {firewallId}
```

This PR is currently missing tests. I'll add these in the next couple of days.

## 📷 Preview

`linode-cli linodes ips-list {linodeId}`

linodes ips-list

`linode-cli firewalls rules-list {firewallId}`

firewalls rules-list


amisiorek-akamai:
Are we trying to fix the initial issue of running the command `linode-cli linodes ips-list --suppress-warnings --format 'private' --json 47740084`?

lgarber-akamai:
@amisiorek-akamai So from what I can tell, the JSON output was correct since the intended functionality of `--format` is to retain the top-level scope of the data but to reduce the number of fields in the output. I think the table and delimited (text) outputs weren’t being displayed properly since there was no proper handler for nested lists, causing the CLI to print the raw Python object.

I’d say this PR isn’t really a “fix” but rather a new feature to support properly printing nested object lists :+1:

amisiorek-akamai:
> 

Apologies for the last reply, forgot to check back into this PR. Ah I see, my other question is whether this PR addresses their initial request, which seems to be "Also, note that I'm not trying to get the public IP, I'm just trying to get the single private IP". Is there any way to filter the subtables such that we can only see that `ipv4.private` table?

lgarber-akamai:
> Apologies for the last reply, forgot to check back into this PR. Ah I see, my other question is whether this PR addresses their initial request, which seems to be "Also, note that I'm not trying to get the public IP, I'm just trying to get the single private IP". Is there any way to filter the subtables such that we can only see that `ipv4.private` table?

Good question! I added a `--table` argument that does exactly this:
```bash
linode-cli linodes ips-list --table ipv4.private {instanceId}
```

If they're looking to parse this data out programmatically in the terminal, I think their best bet would be to parse the JSON output using `jq`:
e.g.
```
linode-cli linodes ips-list {instanceId} --json | jq -r '.[0].ipv4.private[0].address'
```
or alternatively, a newline-delimited output:
```
linode-cli linodes ips-list 48955960 --json | jq -r '.[0].ipv4.public[] | [.address] | join("\n")'
```

amisiorek-akamai:
Just checked, new functionality works as expected!

lgarber-akamai:
https://github.com/Dorthu/openapi3/pull/109 has been merged and will go out on Monday

amisiorek-akamai (approved):

linodecli/baked/response.py
```diff
@@ -149,11 +159,22 @@ def _parse_response_model(schema, prefix=None):
```
zliang-akamai, line 159: Can we add the parameter descriptions into the docstring?

linodecli/baked/response.py
```diff
@@ -133,7 +143,7 @@ def get_string(self, model):
         return value
 
 
-def _parse_response_model(schema, prefix=None):
+def _parse_response_model(schema, prefix=None, nested_list_depth=0):
     """
     Recursively parses all properties of this schema to create a flattened set of
     OpenAPIResponseAttr objects that allow the CLI to display this response in a
```
zliang-akamai, line 149: Can we add the parameter descriptions into the docstring below?

zliang-akamai (approved):
The new output looks great on my end!

linodecli/baked/response.py
lgarber-akamai: Good idea! I'll add those before merging

PR #530, opened 2023-10-06
new: Support specifying explicitly empty lists

### LINKED ISSUE #371: [Bug]: Cannot remove axfr_ips from domain
### CLI Version

linode-cli 5.31.0 Built off spec version 4.144.1

### Command

linode-cli domains update <domainID> --axfr_ips []

### Output

no matches found: []

### Expected Behavior

Remove any existing axfr_ips.

### Actual Behavior

error

### Steps to Reproduce

_No response_

lgarber-akamai:
## 📝 Description

This change allows users to specify explicitly empty lists for array arguments using the below syntax:

```bash
linode-cli networking ip-share --linode_id ... --ips []
```

This is useful for API endpoints where an explicit empty list has unique functionality.

Concerns:
- This could affect edge cases where a user could specify `[]` as a valid list value. We should consider adding a spec extension to opt out of this functionality for certain fields.

Resolves #371 

## ✔️ How to Test

```
make testunit
make INTEGRATION_TEST_PATH="networking/test_networking.py::test_share_ipv4_address" testing
```

### DIFF
--- a/linodecli/arg_helpers.py
+++ b/linodecli/arg_helpers.py
@@ -381,7 +381,9 @@ def action_help(cli, command, action):
                 else ""
             )
             nullable_fmt = " (nullable)" if arg.nullable else ""
-            print(f"  --{arg.path}: {is_required}{arg.description}{nullable_fmt}")
+            print(
+                f"  --{arg.path}: {is_required}{arg.description}{nullable_fmt}"
+            )
 
 
 def bake_command(cli, spec_loc):


jriddle-linode (approved):
If there are fields in the API that are labeled Nullable that shouldn't accept `[]` then we should probably change the field and not add a spec extension.

amisiorek-akamai (commented):
Both tests pass locally, just wanted to double check that intended output of issue is `no matches found: []`

lgarber-akamai:
> Both tests pass locally, just wanted to double check that intended output of issue is `no matches found: []`

Could you share which command is triggering this output? The shell might be parsing it in a weird way, in which case you might need to use quotes around the `[]` 👍

amisiorek-akamai:
Yup, that was it, sorry for the delay! Completely dropped the ball on re-reviewing

amisiorek-akamai (approved):

PR #533, opened 2023-10-20
new: Display region_prices in linode types output

yec-akamai:
## 📝 Description

Add `region_prices` as an extra column in the linode types table to show user DC-specific pricing of each linode type.

## ✔️ How to Test

Unit test: `make testunit`

Manual test: run command `linode-cli linodes types`

## 📷 Preview

Old table:
image (1)

New table:
image

### DIFF
--- a/linodecli/overrides.py
+++ b/linodecli/overrides.py
@@ -55,7 +55,10 @@ def handle_types_region_prices_list(
         return True
 
     output = Table()
-    headers = sorted(json_data["data"][0].keys() - ["addons"], key=len)
+
+    # To ensure the order of the headers and make sure we have region_prices as the last column
+    headers = sorted(json_data["data"][0].keys() - ["addons", "price", "region_prices"], key=len)
+    headers += ["price.hourly", "price.monthly", "region_prices"]
     region_price_sub_headers = ["id", "hourly", "monthly"]
 
     for header in headers:
@@ -77,17 +80,9 @@ def handle_types_region_prices_list(
                     sub_table.add_row(*region_price_row)
                 row += [sub_table]
 
-            elif h == "price":
-                sub_table = Table()
-                for header in ["hourly", "monthly"]:
-                    sub_table.add_column(header, justify="center")
-                sub_table.add_row(
-                    *[
-                        Align(str(linode[h]["hourly"]), align="left"),
-                        Align(str(linode[h]["monthly"]), align="left"),
-                    ]
-                )
-                row += [sub_table]
+            elif h == "price.hourly" or h == "price.monthly":
+                price_headers = h.split(".")
+                row += [Align(str(linode[price_headers[0]][price_headers[1]]), align="left")]
 
             else:
                 row += [Align(str(linode[h]), align="left")]
--- a/linodecli/overrides.py
+++ b/linodecli/overrides.py
@@ -47,9 +47,18 @@ def handle_domains_zone_file(operation, output_handler, json_data) -> bool:
 def handle_types_region_prices_list(
     operation, output_handler, json_data
 ) -> bool:
-    # pylint: disable=unused-argument
     """
     Override the output of 'linode-cli linodes types' to display regional pricing.
+    """
+    return linode_types_with_region_prices(operation, output_handler, json_data)
+
+
+def linode_types_with_region_prices(
+    operation, output_handler, json_data
+) -> bool:
+    # pylint: disable=unused-argument
+    """
+    Parse and reformat linode types output with region prices.
     """
     if len(json_data["data"]) < 1:
         return True
@@ -62,7 +71,6 @@ def handle_types_region_prices_list(
         key=len,
     )
     headers += ["price.hourly", "price.monthly", "region_prices"]
-    region_price_sub_headers = ["id", "hourly", "monthly"]
 
     for header in headers:
         output.add_column(header, justify="center")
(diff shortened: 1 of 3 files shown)


linodecli/overrides.py
```diff
@@ -37,3 +41,65 @@ def handle_domains_zone_file(operation, output_handler, json_data) -> bool:
...
+                sub_table = Table()
+                for header in ["hourly", "monthly"]:
+                    sub_table.add_column(header, justify="center")
+                sub_table.add_row(
+                    *[
+                        Align(str(linode[h]["hourly"]), align="left"),
+                        Align(str(linode[h]["monthly"]), align="left"),
+                    ]
+                )
+                row += [sub_table]
```
lgarber-akamai, line 90: Given there will always only be one entry for `price.hourly` and `price.monthly`, maybe they could become their own columns :eyes:

linodecli/overrides.py
```diff
@@ -37,3 +41,65 @@ def handle_domains_zone_file(operation, output_handler, json_data) -> bool:
...
+) -> bool:
+    # pylint: disable=unused-argument
+    """
+    Override the output of 'linode-cli linodes types' to display regional pricing.
+    """
+    if len(json_data["data"]) < 1:
+        return True
+
+    output = Table()
+    headers = sorted(json_data["data"][0].keys() - ["addons"], key=len)
```
lgarber-akamai, line 58: To make things a bit more readable, we might want to add some logic here to ensure `region_prices` is the last column

linodecli/overrides.py
yec-akamai: Sounds good. I put them into separate columns.

linodecli/overrides.py
```diff
@@ -37,3 +41,60 @@ def handle_domains_zone_file(operation, output_handler, json_data) -> bool:
...
+            if h == "region_prices":
+                sub_table = Table()
+                for header in region_price_sub_headers:
+                    sub_table.add_column(header, justify="center")
+                for region_price in linode[h]:
+                    region_price_row = []
+                    for header in region_price_sub_headers:
+                        region_price_row += (
+                            Align(str(region_price[header]), align="left"),
+                        )
```
zliang-akamai, line 79:
```suggestion
                    region_price_row = [
                        Align(str(region_price[header]), align="left")
                        for header in region_price_sub_headers
                    ]
```

linodecli/overrides.py
```diff
@@ -37,3 +41,60 @@ def handle_domains_zone_file(operation, output_handler, json_data) -> bool:
...
+                    sub_table.add_row(*region_price_row)
+                row += [sub_table]
```
zliang-akamai, line 81:
```suggestion
                row.append(sub_table)
```
I think it may be more efficient and readable to directly append it without making a new list object.

linodecli/overrides.py
```diff
@@ -37,3 +41,60 @@ def handle_domains_zone_file(operation, output_handler, json_data) -> bool:
...
+
+            elif h == "price.hourly" or h == "price.monthly":
```
lgarber-akamai, line 83: Just a small suggestion to make the linter happy, otherwise LGTM!
```suggestion
            elif h in ("price.hourly", "price.monthly"):
```

linodecli/overrides.py
yec-akamai: Use `append` will introduce rich issue and not able to render the table correctly:
`rich.errors.NotRenderableError: unable to render list; a string or other renderable object is required
`

linodecli/overrides.py
```diff
@@ -37,3 +41,67 @@ def handle_domains_zone_file(operation, output_handler, json_data) -> bool:
     """
     print("\n".join(json_data["zone_file"]))
     return False
+
+
+@output_override("linodes", "types", OutputMode.table)
+def handle_types_region_prices_list(
+    operation, output_handler, json_data
+) -> bool:
```
lgarber-akamai, line 49: We may want to move this code out into a separate function so we can partially re-use it for ASCII and markdown outputs :+1:

linodecli/overrides.py
```diff
@@ -37,3 +41,67 @@ def handle_domains_zone_file(operation, output_handler, json_data) -> bool:
...
+
+            elif h in ("price.hourly", "price.monthly"):
+                price_headers = h.split(".")
+                row += [
+                    Align(
+                        str(linode[price_headers[0]][price_headers[1]]),
+                        align="left",
+                    )
+                ]
+
```
lgarber-akamai, line 93: Optional suggestion, feel free to ignore if it adds too much complexity :slightly_smiling_face: :

It might make sense to break out these special cases into their own functions.

e.g.

```python
def format_region_prices(data: Dict[str, Any]) -> any:
    subheaders = ["id", "hourly", "monthly"]

    sub_table = Table()

    for header in subheaders:
        sub_table.add_column(header, justify="center")

    for region_price in data:
        region_price_row = (
            Align(str(region_price[header]), align="left")
            for header in subheaders
        )
        sub_table.add_row(*region_price_row)

    return sub_table
```

tests/unit/test_overrides.py
```diff
@@ -56,3 +56,59 @@ def patch_func(*a):
...
+        stdout_buf = io.StringIO()
+
+        with contextlib.redirect_stdout(stdout_buf):
+            list_operation_for_overrides_test.process_response_json(
+                response_json, mock_cli.output_handler
+            )
+
+        rows = stdout_buf.getvalue().split("\n")
+        # assert that the overridden table has the new columns
+        assert len(rows[1].split("┃")) == 15
```
lgarber-akamai, line 114: Looks like the table characters might be encoded differently on Windows which is causing the GHA tests to fail. 

As a somewhat ugly workaround, you could try resolving the first character in the row and using it to split the table.

e.g.
```python
assert len(rows[1].split(rows[1][0])) == 15
```

lgarber-akamai (changes requested):
The implementation looks good! Just one small test issue

ykim-1 (approved):
Looks good! Everything works well locally

PR #535, opened 2023-10-24
Drop Python 3.7 Support and Remove Usage of `pkg_resources`

zliang-akamai:
## 📝 Description

`pkg_resources` was deprecated and removed in Python 3.12.

## ✔️ How to Test

You can insert a `print(VERSION)` to right after where the `VERSION` is defined. Then run any command with the Linode CLI, and you will be able to see the output.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -100,6 +100,6 @@ setup(
         ]
     },
     data_files=get_baked_files(),
-    python_requires=">=3.7",
+    python_requires=">=3.8",
     include_package_data=True,
 )
--- a/.pylintrc
+++ b/.pylintrc
@@ -307,8 +307,8 @@ min-public-methods=2
 [EXCEPTIONS]
 
 # Exceptions that will emit a warning when caught.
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -100,6 +100,6 @@ def bake_version(v):
         ]
     },
     data_files=get_baked_files(),
-    python_requires=">=3.7",
+    python_requires=">=3.8",
```
yec-akamai, line 103: Should we add a notice somewhere saying that we are dropping support of python 3.7?

yec-akamai (approved):
Nice work! The CI tests are passing now.

amisiorek-akamai (approved):
I agree with Ye, maybe just something in the release notes?

setup.py
zliang-akamai: I think we can put that into the release note.

lgarber-akamai (approved):
Works as expected after installing Python 3.12 and running `python3 -m ensurepip --upgrade` 🎉

PR #537, opened 2023-10-26
removing slash command integration test trigger

vshanthe:
## 📝 Description
Remove slash-command-dispatch action for PR integration test triggering in the repo

The workflow related to slash-command-dispatch is removed and switched to GUI trigger

## 📷 Preview


amisiorek-akamai:
Just for my own sake, why are we removing this workflow in the first place?

vshanthe:
> Just for my own sake, why are we removing this workflow in the first place?

> Some context info(provided by Zhiwei ) is that we used to use this command, but it later requires a GitHub token with org:read scope to run caused by a GitHub API change. We determined that it's not very safe to give a third-party action a token with that scope, so we switched to GUI trigger.

amisiorek-akamai (approved):

jriddle-linode (approved):

zliang-akamai:
Hi @vshanthe, the current protection rules would prevent unsigned commits being merged into the dev branch.

Here is a GitHub doc for signing commits.
https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

PR #539, opened 2023-10-30
test: fix E2E Tests, improve test fixture naming/teardown, label creation 

ykim-1:
## 📝 Description

Make several improvements to integration tests. Here are the requested changes:

Linode Upgrade:
Upgrade Linode generation from g5 to g6-nanode-1 to reduce backend resource consumption.
Update the test environment to use a newer Linode image (e.g., Debian 10 or a later version) instead of the old Debian 9.

Fixture Naming:
Shorten fixture names to be more concise and use nouns. For example, change "create_vpc_with_subnet" to "vpc_with_subnet" or "vpc_subnet."

Unique Labels:
Consider using nanoseconds for unique labels in some tests to ensure uniqueness, as multiple resources can be created within a single second.
These improvements will enhance the efficiency, readability, and relevance of our integration tests.

Also made some fix to tests and improvements in fixture teardowns

## ✔️ How to Test

make testint

There are currently some failures related firewall rules but it is addressed in https://jira.linode.com/browse/TPT-2447

## 📷 Preview


amisiorek-akamai (approved):
Works on my end!

tests/integration/conftest.py
```diff
@@ -122,8 +122,8 @@ def _generate_test_files(
 
 # test helper specific to Domains test suite
 @pytest.fixture
-def create_master_domain():
-    timestamp = str(int(time.time()))
+def master_domain():
+    timestamp = str(int(time.time_ns()))
```
zliang-akamai, line 126: `time.time_ns()` is already an integer, so we don't have to cast it.
```suggestion
    timestamp = str(time.time_ns())
```

zliang-akamai:
Looks great! Thanks for putting them together!

zliang-akamai:
Running it on GHA now:
https://github.com/linode/linode-cli/actions/runs/6737245574

zliang-akamai (approved):

PR #540, opened 2023-10-31
Integration test fixs

vshanthe:
## 📝 Description

Below test cases are failing in Test Outcome Sum
- tests.integration.firewalls.test_firewalls_rules/test_list_rules_json_format
- tests.integration.regions.test_plugin_region_table/test_output

Fixed the failing test cases 

## ✔️ How to Test

run all test 
make testint

To run individual module where fixed has been implemented 
make INTEGRATION_TEST_PATH=regions testint 
make INTEGRATION_TEST_PATH=firewalls/test_firewalls_rules.py testint
 

## 📷 Preview


tests/integration/regions/test_plugin_region_table.py
```diff
@@ -11,13 +12,16 @@ def exec_test_command(args: List[str]):
...
-        assert "-" in line
-        assert "✔" in line
-        assert "|" in line
+        check_marks = re.findall("✔", line)
+        hyphens = re.findall("-", line)
+        pipes = re.findall(r"\|", line)
+
+        assert all(mark in line for mark in check_marks)
+        assert all(hyphen in line for hyphen in hyphens)
+        assert all(pipe in line for pipe in pipes)
```
zliang-akamai, line 27: Nice work! I wonder why we have to loop through `check_marks` if it’s all `✔`?

zliang-akamai:
And you can try `make format` to fix the lint

tests/integration/regions/test_plugin_region_table.py
jriddle-linode: The lines are horizontal, i believe this is checking to see if there are checks, hyphens, and pipes on every line which is probably safe until we hit a datacenter capable of every possible feature.

jriddle-linode (approved):
Looks good to me.

zliang-akamai (approved):

tests/integration/regions/test_plugin_region_table.py
zliang-akamai: But it looks like we also to assert mark in the line if we found the mark in the line, which is kind of weird..

zliang-akamai (changes requested):
I found the issue was caused by the program thinks the console columns is limited so it display some empty lines, like:
image

I think we can use the COLUMNS env var to fix this issue. I found this is a good example for it:
```python
import subprocess
import os

# Set the console width to 100
env = os.environ.copy()
env["COLUMNS"] = "100"

process = subprocess.run(
    args,
    stdout=subprocess.PIPE,
    env=env,
)

```

tests/integration/regions/test_plugin_region_table.py
vshanthe: It was taking the column outer line as well for the asserts which was like this - ┡━━━━━━━╇━━━━━━━╇━━━━━╇━━━━━━╇━━━━━╇━━━━╇━━━━━╇━━━━╇━━━━╇━━━━━╇━━━━━━╇━━━━━━━╇━━━━━━╇━━━━━━╇━━━━━━━┩'

vshanthe (commented):
Changes implemented

zliang-akamai:
It's passing now, great work!
https://github.com/linode/linode-cli/actions/runs/6742822588

zliang-akamai (approved):

PR #541, opened 2023-11-01
test: add vpc tests

ykim-1:
## 📝 Description

writing integration tests that cover all VPC actions. Some actions were manually tested and work well.

Linode configs and interfaces are not in scope of this ticket.

Command format: linode-cli vpcs [ACTION]

image

## ✔️ How to Test

pytest tests/integration/vpc/test_vpc.py

## 📷 Preview


amisiorek-akamai:
Getting errors while running integration tests in alpha:
region = get_regions_with_vpcs_capabilties()[0] is generating a list index error.

ykim-1:
> Getting errors while running integration tests in alpha: region = get_regions_with_vpcs_capabilties()[0] is generating a list index error.

Might have to try with something like return `["us-east"]` for `def get_regions_with_vpcs_capabilties():` function when testing on alpha. 

I tested on production with steps below:
```
export LINODE_CLI_TOKEN=mytoken
export LINODE_CLI_API_HOST=api.linode.com                                               
pytest tests/integration/vpc/test_vpc.py                                                

```

tests/integration/conftest.py
```diff
@@ -374,3 +374,124 @@ def create_nodebalancer_with_default_conf():
...
+    regions = regions.split("\n")
+
+    regions_with_vpc = []
+
+    for region in regions:
+        if "VPCs" in region:
+            region_name = region.split()[0]  # Extract the region name
+            regions_with_vpc.append(region_name)
+
+    return regions_with_vpc
```
zliang-akamai, line 497: Can we make it a global fixture and support getting regions with arbitrary capabilities? Similar to what we have in Terraform provider or linodego.

tests/integration/cli/test_help.py
```diff
@@ -13,7 +13,7 @@ def test_help_page_for_non_aliased_actions():
         "API Documentation: https://www.linode.com/docs/api/linode-instances/#linodes-list"
         in output
     )
-    assert "You may filter results with:" in output
+    assert "wrong assertion" in output
```
zliang-akamai, line 16: Is this an intentional change?

tests/integration/conftest.py
```diff
@@ -374,3 +374,124 @@ def create_nodebalancer_with_default_conf():
...
+            [
+                "linode-cli",
+                "vpcs",
+                "create",
+                "--label",
+                label,
+                "--region",
+                region,
+                "--no-headers",
+                "--text",
```
lgarber-akamai, line 396: I wonder if it'd make sense to start using `--json` and parsing the output just to make sure we're not affected by any `--text` output changes.

tests/integration/cli/test_help.py
ykim-1: No I thought I reverted it back. Should be fixed now

tests/integration/conftest.py
ykim-1: That's a good point and it will be good in the long run like you said. I will create few tickets to use `--json` instead of `--text` in the integration test suites, there are quiet a number of functions/fixtures that use `--text` and `--delimiter`

tests/integration/conftest.py
ykim-1: Made the change

tests/integration/conftest.py
```diff
@@ -374,3 +374,124 @@ def create_nodebalancer_with_default_conf():
...
+                "--no-headers",
+                "--delimiter=,",
+            ]
+        )
+        .stdout.decode()
+        .rstrip()
+    )
+
+    yield res, subnet_label
+
```
zliang-akamai, line 470: Can we move the fixtures dedicated for VPC to the vpc directory and only keep global fixtures here?

ykim-1: moved into test file :)

tests/integration/vpc/test_vpc.py
```diff
@@ -0,0 +1,363 @@
...
+                "--no-headers",
+                "--delimiter=,",
+            ]
+        )
+        .stdout.decode()
+        .rstrip()
+    )
+
+    yield res, subnet_label
+
```
zliang-akamai, line 107: Can we have a cleanup after yielding?

tests/integration/vpc/test_vpc.py
```diff
@@ -0,0 +1,363 @@
...
+    )
+
+    yield vpc_id
+
+    delete_target_id(target="vpcs", id=vpc_id)
+
+
+@pytest.fixture
+def test_subnet(test_vpc_wo_subnet):
+    vpc_id = test_vpc_wo_subnet
```
zliang-akamai, line 84: I feel these fixtures can go to the newly created `conftest.py`?

tests/integration/vpc/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+                "--text",
+                "--no-headers",
+                "--delimiter=,",
+            ]
+        )
+        .stdout.decode()
+        .rstrip()
+    )
+
+    yield res, subnet_label
```
zliang-akamai, line 67: Can we cleanup the subnet after yielding it?

tests/integration/vpc/test_vpc.py
```diff
@@ -0,0 +1,363 @@
...
+    delete_target_id,
+    exec_failing_test_command,
+    exec_test_command,
+)
+
+BASE_CMD = ["linode-cli", "vpcs"]
+
+
+@pytest.fixture
+def test_vpc_wo_subnet():
```
zliang-akamai, line 17: Can we place it in the newly created `conftest.py`?

tests/integration/vpc/conftest.py
ykim-1: Gets deleted with the vpc from `test_vpc_wo_subnet` when tests run, it will throw subnet does not exist error if the instance is deleted before

zliang-akamai (approved):

lgarber-akamai (approved):
Nice work! All added tests pass after merging from dev, installing the CLI from [this spec PR](https://github.com/linode/linode-api-docs/pull/858/files), and running `make INTEGRATION_TEST_PATH=vpc testint`.

We may want to wait to merge this PR until the spec PR is merged, or alternatively we could merge this PR into a feature branch.

PR #544, opened 2023-11-03
Fix `boto3` missing error message in obj extension

zliang-akamai:
## 📝 Description

We imported boto3 in `objects.py` which skips the boto3 check logic `__init__.py`. Added `try ... except ...` block to ignore the exception and let the `call` function in `__init__.py` to print the error message.

## ✔️ How to Test
`make install`
`python3 -m pip uninstall boto3 -y`
`linode obj ls`

GHA run passed: https://github.com/linode/linode-cli/actions/runs/6750190271

### DIFF
--- a/linodecli/plugins/obj/objects.py
+++ b/linodecli/plugins/obj/objects.py
@@ -7,8 +7,13 @@ from argparse import ArgumentParser
 from pathlib import Path
 from typing import List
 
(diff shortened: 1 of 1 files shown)


ykim-1 (approved):
GJ!, works well locally

jcallahan-akamai (approved):
Thanks for the quick fix! It works as expected locally.

PR #546, opened 2023-11-07
Replace `distutils` with `packaging` for Py 3.12

zliang-akamai:
## 📝 Description

`distutils` has been removed in Python 3.12, and we need to replace it with `packaging` in order to get the release workflow works again.

This is targeting `dev` branch intentionally because other code in `dev` branch is needed to get everything working with Python 3.12.

## ✔️ How to Test

1. Install both Python 3.11 and 3.12.
2. `export LINODE_CLI_VERSION=v1.2.3.dev` 
OR
`export LINODE_CLI_VERSION=v1.2.3`
4. Then these three ways should output the same result of version. 
`python3.12 version`
`python3.11 version`
`python3.11 version` (in main branch before merging this PR)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools", "wheel", "packaging"]
 build-backend = "setuptools.build_meta"
 
 [tool.isort]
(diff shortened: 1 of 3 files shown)


lgarber-akamai (approved):
LGTM!

vshanthe (approved):

PR #550, opened 2023-11-16
change loading bar size

jriddle-linode:
## 📝 Description

Change the length of the loading bar based on width of terminal.

image

## ✔️ How to Test

```bash
lin obj mb bucketiquette
lin obj put file.txt bucketiquette
```


lgarber-akamai (approved):
Works well on my end! I wonder if there's a way we can unit test this

vshanthe (approved):

PR #552, opened 2023-11-27
fix: obj plugin error with invalid cluster

jriddle-linode:
## 📝 Description

When using the command with a invalid region it would print out a large stack trace error, to fix this I added a list of valid clusters generated from the API.

## ✔️ How to Test

Use the command with a invalid cluster ID
```bash
lin obj --cluster us-west
```

### DIFF
--- a/linodecli/plugins/obj/__init__.py
+++ b/linodecli/plugins/obj/__init__.py
@@ -451,11 +451,19 @@ def print_help(parser: ArgumentParser):
     print("See --help for individual commands for more information")
 
 
-def get_obj_args_parser():
+def get_obj_args_parser(client: CLI):
     """
     Initialize and return the argument parser for the obj plug-in.
     """
     parser = inherit_plugin_args(ArgumentParser(PLUGIN_BASE, add_help=False))
+    clusters = [
+        c["id"]
+        for c in _do_get_request(  # pylint: disable=protected-access
+            client.config.base_url,
+            "/object-storage/clusters",
+            token=client.config.get_value("token"),
+        )["data"]
+    ]
 
     parser.add_argument(
         "command",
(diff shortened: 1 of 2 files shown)


zliang-akamai:
Unit tests are failing?

lgarber-akamai (changes requested):
Just one small change, otherwise this looks good!

amisiorek-akamai (approved):
Tested locally, works as expected!

linodecli/plugins/obj/__init__.py
```diff
@@ -527,7 +528,15 @@ def call(
 
         sys.exit(2)  # requirements not met - we can't go on
 
-    parser = get_obj_args_parser()
+    clusters = [
+        c["id"]
+        for c in _do_get_request(  # pylint: disable=protected-access
+            context.client.config.base_url,
+            "/object-storage/clusters",
+            token=context.client.config.get_value("token"),
```
zliang-akamai, line 536: This doesn't seem to be working with `LINODE_CLI_TOKEN` environment variable.

```
linode-cli % lin obj --cluster us-west --help
Could not contact https://api.linode.com/v4beta/object-storage/clusters - Error: 401
```

zliang-akamai: This seems to be a separated issue that's also impacting `linode obj configure`. But usage of `--cluster` is much more common than `configure`.

zliang-akamai: I put up a fix here: https://github.com/jriddle-linode/linode-cli/pull/2/
Do you think that would work?

jriddle-linode: Good catch, merged changes, thank you.

zliang-akamai (approved):
Tested locally and worked well! Nice work!

PR #558, opened 2023-12-12
Creating Metadata Plugin

amisiorek-akamai:
## 📝 Description

This PR adds a plug-in to the CLI that allows users to use the Metadata Service API easily. Note: This plug-in will not be functional until py-metadata is publicly available.

## ✔️ How to Test

**Running Unit Tests**

Note: No tests will be functional _unless_ py-metadata is installed on your machine.

`make testunit`

**Manual Testing**

1. Enter Linode instance named _amisiorek-metadata-cli_ as linodedx user using SSH key (if not authorized, message me and I will add your key)

2. Run `source venv/bin/activate` to enter venv

3. Run the following commands to see the plug-in in action:

`linode-cli metadata instance`
`linode-cli metadata sshkeys`
`linode-cli metadata networking`
`linode-cli metadata user-data`
`linode-cli metadata --help`

## 📷 Preview

c44f5e59-1970-4003-853c-317b562ff88a

### DIFF
--- a/linodecli/plugins/metadata.py
+++ b/linodecli/plugins/metadata.py
@@ -185,14 +185,13 @@ def call(args, context):
         sys.exit(0)
 
     # make a client, but only if we weren't printing help and endpoint is valid
-    if not "--help" in args:
+    if "--help" not in args:
         try:
             client = MetadataClient()
         except ConnectTimeout:
-            print(
+            raise ConnectionError(
                 "Can't access Metadata service. Please verify that you are inside a Linode."
             )
-            sys.exit(0)
     else:
         print_help(parser)
         sys.exit(0)
--- a/linodecli/plugins/metadata.py
+++ b/linodecli/plugins/metadata.py
@@ -1,0 +1,146 @@
+"""
+This plugin allows users to access the metadata service while in a Linode.
+
+Usage:
+
+   linode-cli metadata [ENDPOINT]
+"""
+
+import argparse
+import sys
+
+from linode_metadata import MetadataClient
+from linode_metadata.objects.error import ApiError
+from linode_metadata.objects.instance import ResponseBase
+from requests import ConnectTimeout
+from rich import print as rprint
+from rich.table import Table
+
+PLUGIN_BASE = "linode-cli metadata"
+
+
+def process_sub_columns(subcolumn: ResponseBase, table: Table, values_row):
+    for key, value in vars(subcolumn).items():
+        if isinstance(value, ResponseBase):
+            process_sub_columns(value, table, values_row)
+        else:
+            table.add_column(key)
+            values_row.append(str(value))
+
+
+def print_metadata_table(data):
+    attributes = vars(data)
+    values_row = []
+
+    table = Table()
+
+    for key, value in attributes.items():
+        if isinstance(value, ResponseBase):
+            process_sub_columns(value, table, values_row)
+        else:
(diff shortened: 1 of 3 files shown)


lgarber-akamai (changes requested):
Excellent work so far, everything seems to work as expected! 

To make the output a bit more parsable, it might be nice to support a `--json` flag to support raw JSON outputs. This may make sense to split out into a separate ticket/PR since it might require some additional work on the py-metadata end.

linodecli/plugins/metadata.py
```diff
@@ -0,0 +1,203 @@
...
+
+    parser = get_metadata_parser()
+    parsed, args = parser.parse_known_args(args)
+
+    if not parsed.endpoint in COMMAND_MAP or len(args) != 0:
+        print_help(parser)
+        sys.exit(0)
+
+    # make a client, but only if we weren't printing help and endpoint is valid
+    if not "--help" in args:
```
lgarber-akamai, line 188:
```suggestion
    if "--help" not in args:
```

linodecli/plugins/metadata.py
```diff
@@ -0,0 +1,203 @@
...
+        try:
+            client = MetadataClient()
+        except ConnectTimeout:
+            print(
+                "Can't access Metadata service. Please verify that you are inside a Linode."
+            )
```
lgarber-akamai, line 194: Given this is an error, it might make sense to print this to stderr:
```suggestion
            print(
                "Can't access Metadata service. Please verify that you are inside a Linode.",
                file=sys.stderr
            )
```

Alternatively, you could raise an error here (e.g. `raise ConnectionError("Can't access Metadata service. Please verify that you are inside a Linode."`)

linodecli/plugins/metadata.py
```diff
@@ -0,0 +1,203 @@
...
+            sys.exit(0)
```
lgarber-akamai, line 195: Since this is a failure case, we should probably return a non-zero status code 👍

linodecli/plugins/metadata.py
amisiorek-akamai: Just out of curiosity, is there a logical difference between these two? I didn't even realize you could put the `not` on either side 🤔

lgarber-akamai: Good question, as far as I know there isn't a logic difference between the two and `not in` is just a bit more Pythonic.

linodecli/plugins/metadata.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+def print_ssh_keys_table(data):
+    """
+    Prints the table that contains information about the SSH keys for the current instance
+    """
+    table = Table(show_lines=True)
+
+    table.add_column("ssh keys")
+    for key in data.users.root:
```
lgarber-akamai, line 61: It looks like `root` will be None if no authorized keys are configured for the root user, so we'll need a None check here to prevent an error.

Side note: It looks like this endpoint will also return keys for users not configured by the API, so we may need to update py/go-metadata accordingly 👍

lgarber-akamai (changes requested):
Looks great! Just one more thing and this should be good to go 🙂

lgarber-akamai (approved):
LGTM, excellent work!

zliang-akamai (approved):
Tested and worked well on my end. Nice work!

PR #562, opened 2023-12-21
Fix `--help` for obj commands

zliang-akamai:
## 📝 Description

Currently `linode obj put --help` shows an 401 error when the CLI is configured via `linode configure` (not via env vars). This is because it won't load the token when `--help` is detected. So we should also avoid invoking API that requires a token when `--help` is in the command.

## ✔️ How to Test
- On a clean Linode (or local machine that unset all CLI related env vars), install Linode CLI and configure it via `linode configure`.
- Verify `linode obj put --help` is working fine.


jriddle-linode (approved):
LGTM works locally

vshanthe (approved):

PR #563, opened 2023-12-21
Change string concatenation style to be adjacent string literal

zliang-akamai:
## 📝 Description

Change string concatenation style to be adjacent string literal for improved code readability.

## ✔️ How to Test
Set your Linode token into env var called `LINODE_CLI_TOKEN`, and then run `linode configure`, and verify the string is displayed correctly.

### DIFF
--- a/linodecli/configuration/__init__.py
+++ b/linodecli/configuration/__init__.py
@@ -304,10 +304,10 @@ class CLIConfig:
 
         if ENV_TOKEN_NAME in os.environ:
             print(
-                f"""Using token from {ENV_TOKEN_NAME}.
-Note that no token will be saved in your configuration file.
-    * If you lose or remove {ENV_TOKEN_NAME}.
-    * All profiles will use {ENV_TOKEN_NAME}."""
+                f"Using token from {ENV_TOKEN_NAME}.\n"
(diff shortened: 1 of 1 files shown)


ykim-1 (approved):

linodecli/configuration/__init__.py
```diff
@@ -304,21 +304,20 @@ def configure(
...
         if ENV_TOKEN_NAME in os.environ:
             print(
-                f"""Using token from {ENV_TOKEN_NAME}.
-Note that no token will be saved in your configuration file.
-    * If you lose or remove {ENV_TOKEN_NAME}.
-    * All profiles will use {ENV_TOKEN_NAME}."""
+                f"Using token from {ENV_TOKEN_NAME}.\n"
+                "Note that no token will be saved in your configuration file.\n"
+                f"    * If you lose or remove {ENV_TOKEN_NAME}.\n"
+                f"    * All profiles will use {ENV_TOKEN_NAME}.\n"
```
lgarber-akamai, line 310: Just to keep things consistent, I think we can drop the last newline here 👍 

```suggestion
               f"    * All profiles will use {ENV_TOKEN_NAME}."
```

lgarber-akamai (approved):
Looks great and works well locally!

PR #566, opened 2024-01-03
fix: wiki publishing gha

jriddle-linode:
## 📝 Description

Fixes the formatting of the workflow file, tested with ACT but unfortunately secrets dont work so well in ACT, I'm getting the same response I get from remote-release-trigger though so it seems to be progressing past my error now.

## ✔️ How to Test

```bash
act -j publish-wiki
```

### DIFF
--- a/.github/workflows/publish-wiki.yml
+++ b/.github/workflows/publish-wiki.yml
@@ -14,5 +14,5 @@ jobs:
   publish-wiki:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: Andrew-Chen-Wang/github-wiki-action@v4
--- a/.github/workflows/publish-wiki.yml
+++ b/.github/workflows/publish-wiki.yml
@@ -15,4 +15,4 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: Andrew-Chen-Wang/github-wiki-action@v4
+      - uses: Andrew-Chen-Wang/github-wiki-action@50650fccf3a10f741995523cf9708c53cec8912a # pin@v4.4.0
--- a/.github/workflows/publish-wiki.yml
+++ b/.github/workflows/publish-wiki.yml
@@ -5,18 +5,19 @@ on:
     branches: [main]
 
 jobs:
-  build:
+  publish-wiki:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - name: Clone Repository
+        uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-wiki.yml
```diff
@@ -5,18 +5,19 @@ on:
     branches: [main]
 
 jobs:
-  build:
+  publish-wiki:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - name: Clone Repository
+        uses: actions/checkout@v3
```
zliang-akamai, line 12:
```suggestion
        uses: actions/checkout@v4
```
Let's upgrade to v4

.github/workflows/publish-wiki.yml
```diff
@@ -5,18 +5,19 @@ on:
...
 
       - name: Generate App Installation Token
-      id: generate_token
-      uses: tibdex/github-app-token@b62528385c34dbc9f38e5f4225ac829252d1ea92 # pin@v1
-      with:
-        app_id: ${{ secrets.CLI_RELEASE_APP_ID }}
-        private_key: ${{ secrets.CLI_RELEASE_PRIVATE_KEY }}
-        repository: linode/linode-cli
+        id: generate_token
+        uses: tibdex/github-app-token@b62528385c34dbc9f38e5f4225ac829252d1ea92 # pin@v1
```
zliang-akamai, line 16: `tibdex/github-app-token` >= v2 will use newer nodejs 20, do you think we can also upgrade this one?

zliang-akamai (approved):
Looks good with some optional comments!

lgarber-akamai:
I've been getting the following error on the upload step on my fork:
```
fatal: could not read Password for 'https://***@github.com': No such device or address
```

Maybe I'm missing some permissions?

.github/workflows/publish-wiki.yml
```diff
@@ -1,26 +1,18 @@
...
-        repository: linode/linode-cli
-
-      - name: Upload Documentation to Wiki
-        uses: SwiftDocOrg/github-wiki-publish-action@v1
-        with:
-          path: "wiki"
-        env:
-          GH_PERSONAL_ACCESS_TOKEN: ${{ steps.generate_token.outputs.token }}
+      - uses: actions/checkout@v4
+      - uses: Andrew-Chen-Wang/github-wiki-action@v4
```
zliang-akamai, line 18:
```suggestion
      - uses: Andrew-Chen-Wang/github-wiki-action@50650fccf3a10f741995523cf9708c53cec8912a # pin@v4.4.0
```
Let's pin it to be hash just to be safe because it will have write permission to the repo.

zliang-akamai (approved):

PR #571, opened 2024-01-16
fix: Resolve bug that prevented nested dict fields in object lists from being included in request body

lgarber-akamai:
## 📝 Description

This change resolves an issue that caused nested dict fields in object lists (e.g. `--interfaces.ipv4.nat_1_1`) to be excluded from request bodies.

This issue occurred because the logic to traverse API spec request arguments did not preserve whether a nested dict field was under a list. Additionally, the name of the list for a field was resolved using the prefix of a field rather than its top-level list name causing mismatches for deeply nested fields.

## ✔️ How to Test

The following test steps assume you have pulled down this change locally and run `make install`.

### Integration Testing

```bash
make INTEGRATION_TEST_PATH=linodes/test_interfaces.py testint
```

### Unit Testing

```bash
make testunit
```

### Manual Testing

1. Create an arbitrary VPC and VPC subnet:

```bash
export VPC_ID=$(linode-cli vpcs create --label test-vpc --region us-mia --json | jq '.[0].id')
export SUBNET_ID=$(linode-cli vpcs subnet-create --label test-subnet --ipv4 '10.0.1.0/24' ${VPC_ID} --json | jq '.[0].id')
```

2. Create an instance with an interface referencing the new VPC and defining the nested `interfaces.ipv4.nat_1_1` and `interfaces.ipv4.vpc` fields. This command should be run with debug mode enabled (`--debug`).

```bash
linode-cli linodes create \
  --label test-instance \
  --region us-mia \
  --image linode/alpine3.19 \
  --root_pass 'testp4ssw0rd!!!!!!23123' \
  --type g6-nanode-1 \
  --interfaces.primary true --interfaces.purpose vpc --interfaces.subnet_id ${SUBNET_ID} \
  --interfaces.ipv4.nat_1_1 any --interfaces.ipv4.vpc "10.0.1.5" \
  --debug
```

3. Observe that the API request contains the following structure for the interfaces field:

```bash
"interfaces": [{"purpose": "vpc", "primary": true, "subnet_id": 12345, "ipv4": {"vpc": "10.0.1.5", "nat_1_1": "any"}}]
```

### DIFF
--- a/linodecli/baked/request.py
+++ b/linodecli/baked/request.py
@@ -9,7 +9,7 @@ class OpenAPIRequestArg:
     """
 
     def __init__(
-        self, name, schema, required, prefix=None, list_item=False
+        self, name, schema, required, prefix=None, list_parent=None
     ):  # pylint: disable=too-many-arguments
         """
         Parses a single Schema node into a argument the CLI can use when making
@@ -60,7 +60,12 @@ class OpenAPIRequestArg:
         self.item_type = None
 
         #: Whether the argument is a field in a nested list.
-        self.list_item = list_item
+        self.list_item = list_parent is not None
+
+        #: The name of the list this argument falls under.
+        #: This allows nested dictionaries to be specified in lists of objects.
+        #: e.g. --interfaces.ipv4.nat_1_1
+        self.list_parent = list_parent
 
         #: The path of the path element in the schema.
         self.prefix = prefix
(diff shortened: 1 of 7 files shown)


linodecli/baked/request.py
```diff
@@ -9,7 +9,7 @@ class OpenAPIRequestArg:
     """
 
     def __init__(
-        self, name, schema, required, prefix=None, list_item=False
```
lgarber-akamai, line 12: I dropped `list_item` as an argument here since it's made redundant by checking whether `list_parent` is None.

zliang-akamai (approved):
LGTM! Verified that the `ipv4` field is included in the request body when creating a Linode with explicit interface definitions with CLI.

ykim-1 (approved):
Looks good, nice work!

PR #573, opened 2024-01-17
Fix vpc tests by removing `linodes` column

zliang-akamai:
## 📝 Description

Due to the API schema change,  `linodes` column is no longer shown in CLI text mode output. Removing it in this PR

## ✔️ How to Test
`make INTEGRATION_TEST_PATH="vpc" testint`

### DIFF
--- a/tests/integration/vpc/test_vpc.py
+++ b/tests/integration/vpc/test_vpc.py
@@ -84,14 +84,15 @@ def test_list_subnets(test_vpc_w_subnet):
 
     lines = res.splitlines()
 
-    headers = ["id", "label", "ipv4", "linodes"]
+    headers = ["id", "label", "ipv4"]
 
     for header in headers:
         assert header in lines[0]
 
     for line in lines[1:]:
+        print(line)
         assert re.match(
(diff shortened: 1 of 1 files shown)


tests/integration/vpc/test_vpc.py
```diff
@@ -84,14 +84,15 @@ def test_list_subnets(test_vpc_w_subnet):
...
     lines = res.splitlines()
 
-    headers = ["id", "label", "ipv4", "linodes"]
+    headers = ["id", "label", "ipv4"]
 
     for header in headers:
         assert header in lines[0]
 
     for line in lines[1:]:
+        print(line)
```
lgarber-akamai, line 93:
```suggestion
```

lgarber-akamai (approved):
LGTM!

tests/integration/vpc/test_vpc.py
zliang-akamai: Oops

ykim-1 (approved):
Thanks for investigating and fixing the tests

PR #578, opened 2024-02-09
new: Support specifying complex list arguments as JSON

lgarber-akamai:
## 📝 Description

This change adds support for specifying complex lists (e.g. `interfaces`) as a single JSON string rather than a series of arguments. This should improve the UX when writing scripts that may use these complex structures.

This pull request also makes various changes to generalize parent/child logic rather than explicitly relating it to lists.

This PR originally included improvements to the command help pages but I decided to break it out into a separate PR because of how big this PR was getting 🙂 

## ✔️ How to Test

The following test steps assume you have pulled down this PR and run `make install`.

### Integration Testing

```bash
make testint
```

### Unit Testing

```bash
make testunit
```

#### Complex Lists as JSON

1. Attempt to create a Linode with the `interfaces` argument defined as JSON:

```bash
linode-cli linodes create \
  --label test-instance \
  --region us-mia \
  --image linode/alpine3.19 \
  --root_pass 'testp4ssw0rd!!!!!!23123' \
  --type g6-nanode-1 \
  --interfaces '[{"purpose": "public"}, {"purpose": "vlan", "label": "test-vlan"}]' \
  --debug
```

2. Ensure the interfaces JSON is properly represented in the debug request body output.
3. Navigate to the instance in Cloud Manager and ensure it has been configured as expected.


linodecli/baked/operation.py
```diff
@@ -463,10 +463,54 @@ def _add_args_post_put(self, parser) -> List[Tuple[str, str]]:
 
         return list_items
 
+    def _validate_parent_child_conflicts(self, parsed: argparse.Namespace):
```
lgarber-akamai, line 466: I originally wanted to implement this validation using argparse's mutually exclusive groups but I couldn't get it to play nicely with multiple compatible fields.

linodecli/arg_helpers.py
```diff
@@ -398,11 +398,24 @@ def action_help(cli, command, action):
...
+            extensions = []
+
+            if arg.format == "json":
+                extensions.append("JSON")
+
+            if arg.nullable:
+                extensions.append("nullable")
+
+            if arg.is_parent:
+                extensions.append("conflicts with children")
```
lgarber-akamai, line 411: I don't know if I like the wording here, any suggestions would be much appreciated!

linodecli/baked/operation.py
```diff
@@ -463,10 +463,55 @@ def _add_args_post_put(self, parser) -> List[Tuple[str, str]]:
...
+            # If the parent isn't defined, we can
+            # skip this one
+            if getattr(parsed, parent) is None:
+                continue
+
+            if parent not in conflicts:
+                conflicts[parent] = []
+
+            # We found a conflict
+            conflicts[parent].append(arg)
```
zliang-akamai, line 497: We might be able to use [`defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict) for `conflicts` to simplified things. Just an optional suggestion.

zliang-akamai:
I noticed that a VLAN is used in the interfaces of the instance creation command so VPC creation in step 1 is probably not necessary?

linodecli/baked/operation.py
```diff
@@ -463,10 +463,55 @@ def _add_args_post_put(self, parser) -> List[Tuple[str, str]]:
...
+        specified alongside their parent (e.g. --interfaces).
+        """
+        conflicts = {}
+
+        for arg in self.args:
+            parent = arg.parent
+            arg_value = getattr(parsed, arg.path, None)
+
+            if parent is None or arg_value is None:
+                print(arg.path, parent, arg_value)
```
zliang-akamai, line 478: Is this a temp debug output?

lgarber-akamai: Oops, let me fix that

linodecli/baked/operation.py
lgarber-akamai: Nice find!

yec-akamai (commented):
LGTM. However, I'm seeing a couple of similar errors when running the integration tests around put operations, i.e. 
```
>       exec_test_command(BASE_CMD + ["put", str(file_path), bucket_name])
```
This failed when assert the process.retruncode:
```
>       assert process.returncode == 0
```
Not sure if it's related to this PR or a interim issue, but as a heads up. I'm looking into it as well.

lgarber-akamai:
> LGTM. Unit test and manual test passed locally!
> 
> However, I'm seeing a couple of similar errors when running the integration tests around put operations, i.e.
> 
> ```
> >       exec_test_command(BASE_CMD + ["put", str(file_path), bucket_name])
> ```
> 
> This failed when assert the process.retruncode:
> 
> ```
> >       assert process.returncode == 0
> ```
> 
> Not sure if it's related to this PR or a interim issue, but as a heads up. I'm looking into it as well.

Nice find, I'll look into this as well!

zliang-akamai:
Firewall tests are failing on my end.. Are
image
 you able to reproduce it?

lgarber-akamai:
@zliang-akamai Looks like they're failing on my end too, I'll take a quick look

lgarber-akamai:
So it looks like supporting dicts as JSON introduces some weird conflicts with commands that nest complex lists in dicts (e.g. Firewalls). Given the main use-case of this feature is to reduce complexity when specifying complex lists of objects, I'll remove dicts as JSON from this PR 👍

yec-akamai (approved):
Integration test passed! Nice work 🎉

zliang-akamai (approved):
Tested locally. Nice work!

PR #579, opened 2024-02-12
new: Address breaking change in MDS plugin `sshkeys` command

lgarber-akamai:
## 📝 Description

This change addresses a breaking change in the `metadata` plugin that caused the `sshkeys` command to raise an error. This breaking change was made in the Metadata SDK because the MDS can resolve the SSH keys of users configured using cloud-config and we are no longer limited to the root user.

To simply the formatting a bit, each row in the command output now represents a single key/user relationship.

## ✔️ How to Test

The following test steps assume you are on a development Linode, have pulled down this change, and have run `make install`.

### Unit Testing

```
make testunit
```

### Manual Testing

1. Run the following:

```
linode-cli metadata sshkeys
```

2. Ensure the output is similar to the preview below.

## 📷 Preview

linode-cli_–_metadata_py

### DIFF
--- a/linodecli/plugins/metadata.py
+++ b/linodecli/plugins/metadata.py
@@ -57,11 +57,16 @@ def print_ssh_keys_table(data):
     """
     table = Table(show_lines=True)
 
-    table.add_column("ssh keys")
+    table.add_column("user")
+    table.add_column("ssh key")
 
-    if data.users.root is not None:
-        for key in data.users.root:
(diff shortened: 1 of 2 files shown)


yec-akamai (approved):
Nice work! Tests all passed on a development Linode.

jriddle-linode (approved):
LGTM everything is passing.

PR #580, opened 2024-02-12
Add `SKIP_BAKE` Argument to `bake` Target in Makefile

zliang-akamai:
## 📝 Description

This is to allow baking being skipped during `make build` or `make install`

## ✔️ How to Test

- Run `make build` to verified `bake` runs (`python3 -m linodecli bake` in the standard output)
- Run `SKIP_BAKE=1 make build` to verify `bake` doesn't run (`Skipping bake stage` in the standard output)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,12 @@ install: check-prerequisites requirements build
 
 .PHONY: bake
 bake: clean
+ifeq ($(SKIP_BAKE), 1)
+	@echo Skipping bake stage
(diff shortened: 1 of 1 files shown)


lgarber-akamai (approved):
Both `make install` and `make SKIP_BAKE=1 install` seem to work as expected, great work!

yec-akamai (approved):
Works as expected!

PR #587, opened 2024-03-09
Move listing functions of the obj plugin to its own files

zliang-akamai:
## 📝 Description

This is to move some functions of the obj plugin from `__init__.py` to its own file.

## ✔️ How to Test

```bash
make testint INTEGRATION_TEST_PATH=obj
```

```bash
linode obj la --cluster us-mia-1
linode obj ls --cluster us-mia-1
linode obj ls YOUR_BUCKET_NAME --cluster us-mia-1
```

### DIFF
--- a/linodecli/plugins/obj/__init__.py
+++ b/linodecli/plugins/obj/__init__.py
@@ -3,10 +3,7 @@
 CLI Plugin for handling OBJ
 """
 import getpass
-import glob
-import math
 import os
-import platform
 import socket
 import sys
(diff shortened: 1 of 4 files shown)


zliang-akamai:
@yec-akamai, yea, it depends on the Linode CLI config you have. If the cluster is not specified in the config file, `--cluster` argument will be needed in the command.

ykim-1 (approved):
LGTM, works locally

PR #594, opened 2024-03-29
ref: Drop `version` script; consolidate version logic in setup.py

lgarber-akamai:
## 📝 Description

This change drops the `version` script from the repo root in favor of directly embedding the logic in `setup.py`. This script was previously necessary to failover to the latest git tag when no version was specified, but this functionality has since been removed.

NOTE: In the future we plan on moving away from the `baked_version` construct but it is out of the scope of this PR. See TPT-2500 for more details 🙂 

## ✔️ How to Test

1. Pull down this PR locally.
2. Run `make install`.
3. Run `linode-cli --version` and ensure the CLI version in the output is `0.0.0`.
4. Run `LINODE_CLI_VERSION=1.2.3 make install`.
5. Run `linode-cli --version` and ensure the CLI version in the output is `1.2.3`.


yec-akamai (approved):
Works well on my end!

zliang-akamai (approved):
Works well on my end. Nice work!

PR #597, opened 2024-04-08
ref: Refactor the `configuration` package for readability

lgarber-akamai:
## 📝 Description

This pull request refactors the `configuration` package for readability, including:

* Moving all logic out of `__init__.py`
* Adding docstrings for all methods and classes
* Adding static typing to all method params and return values
* Applying minor code refactors and formatting changes

## ✔️ How to Test

N/A


linodecli/configuration/helpers.py
```diff
@@ -206,86 +258,20 @@ def _config_get_with_default(
...
+
+    :returns: The value pulled from the config or the default value.
+    :rtype: Any
     """
     return (
         config.get(user, field) if config.has_option(user, field) else default
     )
-
-
-def _handle_no_default_user(self):  # pylint: disable=too-many-branches
```
lgarber-akamai, line 218: `self` in this case referenced the CLIConfig, so I moved it back into the CLIConfig class as a method 👍

linodecli/configuration/__init__.py
```diff
@@ -1,19 +1,17 @@
...
+from .auth import *
 from .auth import (
     _check_full_access,
     _do_get_request,
     _get_token_terminal,
     _get_token_web,
 )
+from .config import *
+from .helpers import *
 from .helpers import (
```
lgarber-akamai, line 15: Exporting these private methods isn't ideal but it's necessary to maintain compatibility with certain other components of the CLI without making large changes

jriddle-linode: If they're necessary can we just make public wrappers or convert them to public?

jriddle-linode (approved):
LGTM. cli still works locally

linodecli/configuration/auth.py
```diff
@@ -8,15 +8,19 @@
...
 
 TOKEN_GENERATION_URL = "https://cloud.linode.com/profile/tokens"
-# This is used for web-based configuration
+
+# The hardcoded OAuth client ID for use in web authentication.
+# This client object exists under an official Linode account.
 OAUTH_CLIENT_ID = "5823b4627e45411d18e9"
-# in the event that we can't load the styled landing page from file, this will
+
+# Un the event that we can't load the styled landing page from file, this will
```
zliang-akamai, line 23: Is `Un` a typo here?

zliang-akamai (approved):
Looks good and works well to me

linodecli/configuration/auth.py
lgarber-akamai: Oops, let me fix that!

PR #605, opened 2024-04-30
integration tests

vshanthe:
## 📝 Description
add test for profile and support features 

## ✔️ How to Test
make INTEGRATION_TEST_PATH=users testint
make INTEGRATION_TEST_PATH=support testint

## 📷 Preview


tests/integration/conftest.py
```diff
@@ -216,10 +216,10 @@ def linode_with_label():
         .rstrip()
     )
 
-    yield result
-
     res_arr = result.split(",")
     linode_id = res_arr[4]
+    yield result, linode_id
+
```
ykim-1, line 222: Might be cleaner to make a new fixture within the test file since fixtures in conftest are meant to be generic. This yield might be too specific to be used by other test suites

ykim-1:
Rebase will probably make the PR a bit cleaner along with the comment I left regarding fixture in conftest. Otherwise, LGTM!

tests/integration/conftest.py
vshanthe: i have altered the other files where respective fixture is used that way we can reuse the same method any how i will create on more in the conftest with linode create since it would generic in may test if we needs to optimize

tests/integration/support/test_support.py
```diff
@@ -0,0 +1,100 @@
+import pytest
+
+from tests.integration.helpers import assert_headers_in_lines, exec_test_command
+
+BASE_CMD = ["linode-cli", "tickets"]
+
+
+# this will create a support ticket on your account
```
ykim-1, line 8: Can we have a cleanup/teardown for the tickets that are created?

tests/integration/support/test_support.py
```diff
@@ -0,0 +1,100 @@
...
+        .stdout.decode()
+        .rstrip()
+    )
+    lines = res.splitlines()
+    headers = ["summary", "opened_by", "opened"]
+    assert_headers_in_lines(headers, lines)
+
+
+@pytest.fixture
+def get_ticket_id():
```
ykim-1, line 38: Small thing, it is best practice to name the fixtures with descriptive names or prefix the fixtures instead of starting from a verb. So in this case, I think the something like `ticket_id` would be appropriate.

ykim-1:
LGTM, just small fixes I mentioned above

ykim-1 (approved):

tests/integration/support/test_support.py
vshanthe: tickets created won't be able to delete form command, while clean-up we need to mark them close 
 we have  command - "linode-cli tickets close ticket_id" it gives request failed error

tests/integration/support/test_support.py
zliang-akamai, line 1: I would hesitate to create a real support ticket just for testing cli. Can we discuss it as a team and make decision about it? For now I would suggest skipping all tests that requires creation of a support ticket.

tests/integration/support/test_support.py
```diff
@@ -0,0 +1,100 @@
...
+        .stdout.decode()
+        .rstrip()
+        .split(",")
+    )
+    first_id = ticket_ids[0]
+    yield first_id
+
+
+def test_tickets_view(get_ticket_id):
+    ticket_id = get_ticket_id
```
zliang-akamai, line 61: `get_ticket_id` has been updated to `ticket_id` apparently

tests/integration/conftest.py
```diff
@@ -450,3 +449,41 @@ def pytest_configure(config):
...
+            ]
+        )
+        .stdout.decode()
+        .rstrip()
+    )
+
+    res_arr = result.split(",")
+    linode_id = res_arr[4]
+    yield linode_id
+    delete_target_id(target="linodes", id=linode_id)
```
zliang-akamai, line 489: Any usage for this function? I think `create_linode` has been implemented in somewhere else?

vshanthe: it was defined in conftest with - def linode_with_label(): and using of i.e 
yield result, linode_id so that new function won't needed 
youjung suggested to have separate fixture and 
one more function is implemented in linode_types with has limitation for package

tests/integration/support/test_support.py
vshanthe: Yes sure creation can be skipped for now

zliang-akamai (approved):
LGTM! Nice work!

tests/integration/conftest.py
```diff
@@ -450,3 +449,41 @@ def pytest_configure(config):
     config.addinivalue_line(
         "markers", "smoke: mark test as part of smoke test suite"
     )
+
+
+@pytest.fixture
+def create_linode():
```
ykim-1, line 455: Can we use a different fixture name?
Best practice is to use descriptive name or use a prefix noun instead of verb.

PR #610, opened 2024-05-02
chore: fix python version ref and update actions to read from `.python-version`

chenrui333:
## 📝 Description

```
pyenv: version `linode-cli' is not installed (set by /xxx/linode-cli/.python-version)
make: *** [black] Error 1
```


jriddle-linode:
We're using `.python-version` to maintain virtual environments not just set the python version. If we change this to just use a version number any other repos using that version will have dependency cross over.

chenrui333:
oh, that is interesting, the reason I made such change is because it does not work on my machine :)

jriddle-linode:
Yes we have a internal tool that sets up all the repos and environments for us, we only recently switched to using `pyenv` so we hadn't updated the Contributing docs or anything to call out that setup step.

`pyenv virtualenv 3.8.10 linode-cli` should fix it for you, other versions also work.

zliang-akamai:
Thank you for bringing this up. I am closing this up. Let us know if @jriddle-linode's solution still doesn't work on your end or if you have any comment or question.

PR #617, opened 2024-05-29
fix: Resolve cell rendering error in OBJ plugin

lgarber-akamai:
## 📝 Description

This change resolves an issue that caused the following error to be raised when attempting to render cells with an integer value in the OBJ plugin:

```
rich.errors.NotRenderableError: unable to render int; a string or other renderable object is required
```

## ✔️ How to Test

The following test steps assume you have pulled down this PR and run `make install`.

### Integration Testing

```
make MODULE=obj testint
```

### Manual Testing

N/A

### DIFF
--- a/linodecli/plugins/obj/helpers.py
+++ b/linodecli/plugins/obj/helpers.py
@@ -125,7 +125,7 @@ def _borderless_table(data):
     """
     tab = Table.grid(padding=(0, 2, 0, 2))
     for row in data:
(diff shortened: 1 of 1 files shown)


zliang-akamai (approved):
Verified that the issue has been resolved. Nice work!

ykim-1 (approved):
Looks good, verified that it is working as expected

PR #622, opened 2024-06-20
Added enum for exit codes

ErikZilber:
## 📝 Description

Added enum for exit codes and modified existing sys.exit calls to use it


linodecli/exit_codes.py
```diff
@@ -0,0 +1,16 @@
+"""
+This is an enumeration of the various exit codes in Linode CLI
+"""
+
+from enum import Enum
+
+class ExitCodes(Enum):
```
lgarber-akamai, line 7: We can use the `IntEnum` base class here to avoid needing to get the underlying value of an enum using `.value` 👍

ErikZilber: Fixed!

lgarber-akamai (approved):
LGTM after linting is fixed, nice work! During the next release we should communicate that some exit codes may have changed since the previous release.

yec-akamai (approved):
LGTM

PR #630, opened 2024-07-31
Change Clusters Choice in OBJ Plugin Config to be Text Input

zliang-akamai:


yec-akamai (approved):
Can we have some tests and description for this PR?

jriddle-linode (approved):
lgtm

PR #635, opened 2024-08-16
Add suppress-version-warning config option

### LINKED ISSUE #582: [Feature]: Config based "--suppress-warnings" or, more specifically, quiet version mismatch warning
### Description (Problem / Annoyance)

Calling `linode-cli` without `--suppress-warnings` will almost always output a version mismatch warning at the rate this library changes. Using `linode-cli` in scripts requires adding this argument to every call; or get noisy version mismatch messages in logs that are not really helpful for automated processes.

### Proposal

This flag setting should be configurable via `./config/linode-cli` or an ENV VAR (or both ideally). 

- this could be for suppressing all warnings, as it is
- OR, specifically suppressing the version mismatch warning may also be desirable (seems to be the only one that is quite noisy)
- OR both, for example (naming tbd)
    - LINODE_CLI_SUPPRESS_WARNINGS (would suppress all)
    - LINODE_CLI_SUPPRESS_VERSION_WARNING (just the version mismatch)

### References

Here is the [code](https://github.com/linode/linode-cli/blob/cef8240bcd2629e7ea781454f72c0711325ee72a/linodecli/api_request.py#L302) where the warning is generated.

[Example usage](https://github.com/linode/linode-cli/blob/cef8240bcd2629e7ea781454f72c0711325ee72a/linodecli/configuration/__init__.py#L264) currently relying on `sys.argv` instead of `CLI.suppress_warnings` ([here](https://github.com/linode/linode-cli/blob/cef8240bcd2629e7ea781454f72c0711325ee72a/linodecli/cli.py#L36))

### Extra

I could potentially help out with this if it is something that the community/maintainers are open to.

### Potential Usage Example

```
linode-cli account view
The API responded with version 4.173.0, which is newer than the CLI's version of 4.170.2.  Please update the CLI to get access to the newest features.  You can update with a simple `pip3 install --upgrade linode-cli`
┌────────────┬───────────┬──────────────────────┬─────────┬────────────────────┐
│ first_name │ last_name │ email                │ balance │ balance_uninvoiced │
├────────────┼───────────┼──────────────────────┼─────────┼────────────────────┤
│ Larry      │ Linode    │ linode@linode.com    │ 0.0     │ 1,000,000.00       │
└────────────┴───────────┴──────────────────────┴─────────┴────────────────────┘


export LINODE_CLI_SUPPRESS_VERSION_WARNING=true 
linode-cli account view
┌────────────┬───────────┬──────────────────────┬─────────┬────────────────────┐
│ first_name │ last_name │ email                │ balance │ balance_uninvoiced │
├────────────┼───────────┼──────────────────────┼─────────┼────────────────────┤
│ Larry      │ Linode    │ linode@linode.com    │ 0.0     │ 1,000,000.00       │
└────────────┴───────────┴──────────────────────┴─────────┴────────────────────┘
```

Or this could be something like this in `linode-cli` config
```
[DEFAULT]
default-user = DEFAULT
region = us-southeast
type = g6-standard-2
suppress-warnings = false
suppress-version-warning = true
```

jriddle-linode:
## 📝 Description

Adds `get_bool('value')` method for configuration options and adds a step in configuration to support `suppress-version-warning`.
Also adds support for `LINODE_CLI_SUPPRESS_VERSION_WARNING` environment variable.

## ✔️ How to Test

1. Install the new version of the CLI
```bash
make install
```
2. Go through the configuration process
```bash
lin configure
```
3. The last question should ask you about suppressing API Version Warnings
4. Use the CLI and see if you view this warning

### Environment variable

1. Install the new CLI or delete the value from your config
2. Use the cli after setting the environment variable
```bash
export LINODE_CLI_SUPPRESS_VERSION_WARNING=true
lin linodes ls
unset $LINODE_CLI_SUPPRESS_VERSION_WARNING
lin linodes ls
```
3. Verify the output doesn't have the error.

```bash
make testunit
```

resolves #582

### DIFF
--- a/linodecli/api_request.py
+++ b/linodecli/api_request.py
@@ -374,11 +374,9 @@ def _attempt_warn_old_version(ctx, result):
                 "with --suppress-warnings",
                 file=sys.stderr,
             )
-        suppress_version_warning = False
-        if ctx.config.get_bool("suppress-version-warning") or os.getenv(
+        suppress_version_warning = ctx.config.get_bool("suppress-version-warning") or os.getenv(
             "LINODE_CLI_SUPPRESS_VERSION_WARNING"
-        ):
-            suppress_version_warning = True
+        )
         if new_version_exists and not suppress_version_warning:
             print(
                 f"The API responded with version {spec_version}, which is newer than "
--- a/linodecli/api_request.py
+++ b/linodecli/api_request.py
@@ -4,6 +4,7 @@ This module is responsible for handling HTTP requests to the Linode API.
 
 import itertools
 import json
+import os
 import sys
 import time
 from typing import Any, Iterable, List, Optional
@@ -373,8 +374,12 @@ def _attempt_warn_old_version(ctx, result):
                 "with --suppress-warnings",
                 file=sys.stderr,
             )
-
-        if new_version_exists:
+        suppress_version_warning = False
+        if ctx.config.get_bool("suppress-version-warning") or os.getenv(
+            "LINODE_CLI_SUPPRESS_VERSION_WARNING"
+        ):
+            suppress_version_warning = True
+        if new_version_exists and not suppress_version_warning:
             print(
                 f"The API responded with version {spec_version}, which is newer than "
                 f"the CLI's version of {ctx.spec_version}.  Please update the CLI to get "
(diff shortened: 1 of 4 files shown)


zliang-akamai (approved):
The functionality works well, nice work!

lgarber-akamai (approved):
This looks great and works well locally, nice work!

PR #638, opened 2024-08-28
new: Break out command help page logic; sort actions in output

lgarber-akamai:
## 📝 Description

This pull request breaks the CLI command help page logic out into a separate function and implements the logic to sort action entries by name. This should improve the readability of certain groups since the order of actions in the spec is not guaranteed to be readable.

Additionally, this PR adds a new test case to validate the output of this help page.

## ✔️ How to Test

The following test steps expect you have pulled down this PR and run `make install`.

### Unit Testing

```shell
make testunit
```

### Manual Testing

1. View the help page for a single CLI command/group:

```shell
linode-cli networking --help
```

2. Ensure all printed actions are sorted by name.

### DIFF
--- a/linodecli/__init__.py
+++ b/linodecli/__init__.py
@@ -27,6 +27,7 @@ from .configuration import ENV_TOKEN_NAME
 from .help_pages import (
     HELP_TOPICS,
     print_help_action,
+    print_help_command_actions,
     print_help_commands,
     print_help_default,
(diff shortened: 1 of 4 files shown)


zliang-akamai (approved):
Works well on my end!

yec-akamai (approved):
Works well on my end!

PR #642, opened 2024-09-10
test: fix unit tests to not overwrite existing cli config file

ykim-1:
## 📝 Description

Running unit tests caused existing config in `~/.config/linode-cli` to be overwritten, this PR fixes this issue by setting the environment variable prior to the test execution so the config file will get written in temporary location. After the execution the environment variable will be reverted to original value.

## ✔️ How to Test

1. unset `LINODE_CLI_TOKEN`
2. `make install`
3. `linode-cli linodes ls` - Expected to fail
4. `make testunit`
5. verify ~/.config/linode-cli is not overwritten

## 📷 Preview

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,9 +66,12 @@ clean:
 	rm -rf dist linode_cli.egg-info build
 
 .PHONY: testunit
-testunit: export LINODE_CLI_TEST_MODE = 1
 testunit:
-	pytest -v tests/unit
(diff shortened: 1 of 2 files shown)


zliang-akamai (approved):
LGTM, works well on my end!

yec-akamai (approved):
Tested and works well on my end!

PR #648, opened 2024-09-26
fix: get-kubeconfig merge into empty lists

merll:
## 📝 Description

This PR fixes the plugin `get-kubeconfig` merging a Kubeconfig from the Linode API with a local kubeconfig file, in the case that all other clusters, contexts, or users have been discarded before. The problem can occur because standard kubeconfig editing tools (e.g. `kubectl config`) set the list entries to `null` instead of an empty list `[]`.

Also the output when adding the `--dry-run` parameter is now showing YAML output (i.e. the expected Kubeconfig update after running the command) instead of a Python dict.

## ✔️ How to Test

1. From an existing Kubeconfig file, remove all contexts using `kubectl config delete-context <context_name>`.
2. Note that the value in this Kubeconfig file now contains a root level value `contexts: null` instead of `contexts: []`.
3. Run `linode-cli get-kubeconfig --label <cluster_label>` to fetch a new config file.
    * Without the PR, the Kubeconfig file is not updated.
    * With the patch, the `contexts` root level item now contains the entry retrieved from the Linode API.
4. Running `linode-cli get-kubeconfig --label <cluster_label> --dry-run`
    * without the patch dumps the string representation of a Python dict which looks somewhat similar to JSON but is not usable
    * with the patch shows the output that would have been written to the Kubeconfig file.

The following unit test has been added to cover the specific case of merging entries into `null` lists:
`pytest tests/unit/test_plugin_kubeconfig.py::test_merge_to_empty_config`

### DIFF
--- a/linodecli/plugins/get-kubeconfig.py
+++ b/linodecli/plugins/get-kubeconfig.py
@@ -90,7 +90,7 @@ def call(args, context):
         else cluster_config
     )
     if parsed.dry_run:
-        print(cluster_config)
+        print(yaml.dump(cluster_config))
     else:
         _dump_config(kubeconfig_path, cluster_config)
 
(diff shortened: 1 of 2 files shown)


zliang-akamai (approved):
Works well for me. Thanks for fixing it up!

ezilber-akamai (approved):
Everything working as expected locally. Thanks for the contribution!

lgarber-akamai (approved):
Looks great, thanks for the contribution!

PR #652, opened 2024-10-02
new: Allow specifying a custom configuration path through environment variable

yec-akamai:
## 📝 Description

User can specify a custom configuration path to replace the default one through a new environment variable `LINODE_CLI_CONFIG`. If this path does not exist yet, we'll create it inexplicably. 

Fixed a behavior change to allow config update look for DEFAULT section. 

### Unit Test:
```
make testunit
```

### Manual Test
1. Pull this PR and install the change locally:
```
make install
```
2. Run the help topic and see the new environment variable is available in the list:
```
linode-cli env-var
```
3. Set a custom config path, i.e.:
```
export LINODE_CLI_CONFIG="~/.test-config/cli-config"
```
4. Reconfigure CLI:
```
linode-cli configure
```
5. Find the reconfigure finished successfully and configuration is written into the new file.
6. Run a command and observe that the allowed default values are used, i.e.: 
```
linode-cli linodes create --debug
```
This command will fail due to missing root_pass, but we can see the configed defaults are passed in the body.
8. Unset the `LINODE_CLI_CONFIG`
9. Reconfigure CLI and find it's successfully written into the default file

### DIFF
--- a/linodecli/configuration/config.py
+++ b/linodecli/configuration/config.py
@@ -290,7 +290,10 @@ class CLIConfig:
         ):
             print(f"User {username} is not configured.")
             sys.exit(ExitCodes.USERNAME_ERROR)
-        if not self.config.has_section(username) or allowed_defaults is None:
+        if (
+            not self.config.has_section(username)
+            and self.config.default_section is None
+        ) or allowed_defaults is None:
             return namespace
 
         warn_dict = {}
--- a/linodecli/configuration/config.py
+++ b/linodecli/configuration/config.py
@@ -335,12 +335,6 @@ class CLIConfig:
         to save values they've set, and is used internally to update the config
         on disk when a new user if configured.
         """
-
-        # Create the config path isf necessary
-        config_path = f"{os.path.expanduser('~')}/.config"
-        if not os.path.exists(config_path):
-            os.makedirs(config_path)
-
         with open(_get_config_path(), "w", encoding="utf-8") as f:
             self.config.write(f)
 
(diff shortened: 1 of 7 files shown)


linodecli/configuration/config.py
```diff
@@ -290,7 +290,10 @@ def update(
         ):
             print(f"User {username} is not configured.")
             sys.exit(ExitCodes.USERNAME_ERROR)
-        if not self.config.has_section(username) or allowed_defaults is None:
+        if (
```
yec-akamai, line 293: Not sure how it used to work, but currently the `DEFAULT` section can't be found by calling `has_section("DEFAULT")`. It will skip update and use the defaults. Add a check to look for the default section under which we put the configured values at the most time.

Luckily we can still use `config.has_option("DEFAULT", key)` to get the value

linodecli/configuration/config.py
```diff
@@ -335,12 +338,6 @@ def write_config(self):
         to save values they've set, and is used internally to update the config
         on disk when a new user if configured.
         """
-
```
yec-akamai, line 338: Moved this logic to `_get_config_path()`.

lgarber-akamai (commented):
Works perfectly in my manual testing, nice work!
Is there any chance a unit test could be added to test this functionality (e.g. using the [tmp_path fixture](https://docs.pytest.org/en/6.2.x/reference.html#std-fixture-tmp_path))?

lgarber-akamai (approved):
Looks great, nice work!

jriddle-linode (approved):
LGTM

PR #653, opened 2024-10-07
Override images replicate output

yec-akamai:
## 📝 Description

Override `images replicate` output to leverage `replicas`. Also improve the table format and make it more readable.

## ✔️ How to Test

1. Pull this PR and install this change:
```
make install
```
2. Replicate an existing image to some regions, i.e.:
```
linode images replicate private/123456 --regions us-iad --regions us-ord --regions us-east
```
3. Observe the command works successfully and the output looks good. An example output:
Untitled

### DIFF
--- a/linodecli/overrides.py
+++ b/linodecli/overrides.py
@@ -4,7 +4,7 @@ This allows us to easily alter per-command outputs, etc. without making
 large changes to the OpenAPI spec.
 """
 
-from typing import Dict
+from typing import Dict, List
 
 from rich import box
 from rich import print as rprint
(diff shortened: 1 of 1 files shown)


ezilber-akamai (approved):
Tested locally and everything works. Nice job!

lgarber-akamai (commented):
Works well in my local testing, nice work!

To make things a bit more readable, is there any chance some of the less-important fields could be excluded from the table view? (vendor, EOL, etc.)

yec-akamai:
> Works well in my local testing, nice work!
> 
> To make things a bit more readable, is there any chance some of the less-important fields could be excluded from the table view? (vendor, EOL, etc.)

This a good point! I excluded fields that have `None` value from the table view.

lgarber-akamai (approved):
Thanks for addressing my feedback!

PR #658, opened 2024-10-17
Sanitize Bearer Token in debug output

ezilber-akamai:
## 📝 Description

Updated debug output to sanitize bearer token

## ✔️ How to Test

Pull down this PR and run `make install`

### Unit Tests
`make testunit`

### Manual Tests
After running `make install`, run any CLI command with the debug flag (i.e. `linode-cli firewalls list --debug`) and verify that the Bearer Token is sanitized in the output.

### DIFF
--- a/linodecli/api_request.py
+++ b/linodecli/api_request.py
@@ -285,7 +285,7 @@ def _print_request_debug_info(method, url, headers, body):
     for k, v in headers.items():
         # If this is the Authorization header, sanitize the token
         if k.lower() == "authorization":
-            v = "Bearer *******************************"
+            v = "Bearer " + "*" * 64
         print(f"> {k}: {v}", file=sys.stderr)
     print("> Body:", file=sys.stderr)
     print(">  ", body or "", file=sys.stderr)
--- a/linodecli/api_request.py
+++ b/linodecli/api_request.py
@@ -283,6 +283,9 @@ def _print_request_debug_info(method, url, headers, body):
     """
     print(f"> {method.__name__.upper()} {url}", file=sys.stderr)
     for k, v in headers.items():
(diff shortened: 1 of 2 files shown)


linodecli/api_request.py
```diff
@@ -283,6 +283,9 @@ def _print_request_debug_info(method, url, headers, body):
     """
     print(f"> {method.__name__.upper()} {url}", file=sys.stderr)
     for k, v in headers.items():
+        # If this is the Authorization header, sanitize the token
+        if k.lower() == "authorization":
+            v = "Bearer *******************************"
         print(f"> {k}: {v}", file=sys.stderr)
```
lgarber-akamai, line 289: Just to make this behavior a bit more clear, could we introduce a new local variable here rather than mutating `v`?

linodecli/api_request.py
```diff
@@ -283,6 +283,9 @@ def _print_request_debug_info(method, url, headers, body):
     """
     print(f"> {method.__name__.upper()} {url}", file=sys.stderr)
     for k, v in headers.items():
+        # If this is the Authorization header, sanitize the token
+        if k.lower() == "authorization":
+            v = "Bearer *******************************"
```
lgarber-akamai, line 288: nitpick: We can use `"*" * 64` to simplify this string literal a tiny bit

lgarber-akamai (approved):
Unit tests are passing locally and everything looks good in my manual testing. Great work!

linodecli/api_request.py
ezilber-akamai: Fixed!

jriddle-linode (approved):
LGTM nice catch.

PR #659, opened 2024-10-21
Simplify and colorize images replicate table

yec-akamai:
## 📝 Description

Simplify the `images replicate` table to only display highly-related fields of a image replication. User can find more information in a single image view.

Add a table theme to colorize the image replicas status.

## ✔️ How to Test

Old view:
image

New view:
image


ezilber-akamai (approved):
Tested locally and output looks correct. This is a much nicer format!

PR #675, opened 2024-11-05
Override placement group view to display members

yec-akamai:
## 📝 Description

Implement output override to display the complex members list in placement group result. 

Override the following commands:

- placement group-view
- placement group-update
- placement assign-linode
- placement unassign-linode

## ✔️ How to Test

1. Pull this PR locally and install the change
```
make install
```
2. Test the above overridden commands to see if it works as expected, for instance:
```
linode-cli placement group-view $pg-id
```
Expected view:
pg

### DIFF
--- a/linodecli/overrides.py
+++ b/linodecli/overrides.py
@@ -76,6 +76,46 @@ def handle_image_replicate(operation, output_handler, json_data) -> bool:
     return image_replicate_output(json_data)
 
 
+@output_override("placement", "group-view", OutputMode.table)
+def handle_placement_group_view(operation, output_handler, json_data) -> bool:
+    # pylint: disable=unused-argument
+    """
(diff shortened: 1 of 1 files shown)


linodecli/overrides.py
```diff
@@ -76,6 +76,46 @@ def handle_image_replicate(operation, output_handler, json_data) -> bool:
     return image_replicate_output(json_data)
 
 
+@output_override("placement", "group-view", OutputMode.table)
```
yec-akamai, line 79: I didn't override the `placement groups-list` output here, because I think the list view is already huge itself if user have multiple PGs, and we can try to keep it simple. User can check the singe group view for members details in this case. Let me know if you have different thoughts!

lgarber-akamai (approved):
The nested table is super clean, nice work!

jriddle-linode (approved):
Looks great!

PR #664, opened 2024-10-25
Allow overriding spec when manually triggering E2E test workflow

vshanthe:
## 📝 Description
Maintainers can manually specify the URL of an OpenAPI spec to use when dispatching the E2E test workflow

**If applicable, include a screenshot or code snippet of this change. Otherwise, please remove this section.**

PR #679, opened 2024-11-13
Allow overriding spec when manually triggering E2E test workflow #664

vshanthe:
## 📝 Description
Maintainers can manually specify the URL of an OpenAPI spec to use when dispatching the E2E test workflow


.github/workflows/e2e-suite.yml
```diff
@@ -95,7 +105,14 @@ jobs:
         run: |
           timestamp=$(date +'%Y%m%d%H%M')
           report_filename="${timestamp}_cli_test_report.xml"
-          make testint TEST_ARGS="--junitxml=${report_filename}" MODULE="${{ inputs.module }}"
+          echo "Running tests, report: ${report_filename}"
+          if [ -n "$OPENAPI_SPEC_URL" ]; then
+            echo "Running tests with OpenAPI spec from: $OPENAPI_SPEC_URL"
+            make testint TEST_ARGS="--junitxml=${report_filename}"
```
ykim-akamai, line 111: Looks like we are running the same command, can we just leave this as it is? 
MODULE parameter was cut out as well

.github/workflows/e2e-suite.yml
```diff
@@ -82,8 +86,14 @@ jobs:
...
 
+      - name: Set OpenAPI Spec URL
+        if: ${{ inputs.openapi_spec_url != '' }}
+        run: |
+          echo "Using OpenAPI Spec URL: ${{ inputs.openapi_spec_url }}"
+          echo "OPENAPI_SPEC_URL=${{ inputs.openapi_spec_url }}" >> $GITHUB_ENV
+
       - name: Install Package
-        run: make install
+        run: make install SPEC="${{ env.OPENAPI_SPEC_URL }}"
```
ykim-akamai, line 96: I think it might be cleaner to just input it directly here. `SPEC="${{ inputs.openapi_spec_url }}"`

.github/workflows/e2e-suite.yml
```diff
@@ -82,8 +86,14 @@ jobs:
           pip install certifi -U && \
           pip install .[obj,dev]
 
+      - name: Set OpenAPI Spec URL
```
ykim-akamai, line 89: Probably don't need this extra step if we input it directly. Refer to comment below

.github/workflows/e2e-suite.yml
```diff
@@ -111,7 +116,14 @@ jobs:
...
           timestamp=$(date +'%Y%m%d%H%M')
           report_filename="${timestamp}_cli_test_report.xml"
-          make testint TEST_ARGS="--junitxml=${report_filename}" MODULE="${{ inputs.module }}"
+          echo "Running tests, report: ${report_filename}"
+          if [ -n "$OPENAPI_SPEC_URL" ]; then
+            echo "Running tests with OpenAPI spec from: $OPENAPI_SPEC_URL"
+            make testint TEST_ARGS="--junitxml=${report_filename}" MODULE="${{ inputs.module }}"
+          else
+                echo "Running tests with default OpenAPI spec"
+                make testint TEST_ARGS="--junitxml=${report_filename}"
```
ykim-akamai, line 125: I think you forgot the MODULE here. I think it would be nice if we just keep it simple and not echo anything in this step. 
e.g. `make testint TEST_ARGS="--junitxml=${report_filename}" MODULE="${{ inputs.module }}"`

Can you confirm one more time with forked run as well after the changes?

vshanthe: Screenshot 2024-11-20 at 9 56 39 PM

ykim-akamai:
Forked run seems to be ok - https://github.com/ykim-akamai/linode-cli/actions/runs/11939572014/job/33280378849

ykim-akamai (approved):
LGTM, nice work!

PR #681, opened 2024-11-14
Exclude parameter values from API request bodies

lgarber-akamai:
## 📝 Description

This pull request fixes an issue where parameter values (e.g. `instanceId` for `/linode/instances/{instanceId}`) would be included in the request body in addition to being used to build the request's URL. This doesn't seem to cause issues on most endpoints but is causing some validation issues against certain DBaaS 2.0 endpoints.

## ✔️ How to Test

### Unit Testing

```
make testunit
```

### Manual Testing

1. Run the following command (NOTE: This is expected to fail because the instance does not exist)

```shell
linode-cli linodes update --label foobar 12345 --debug
```

2. In the debug output, ensure the request URL matches `PUT https://api.linode.com/v4/linode/instances/12345` and the request body does _not_ contain the `linodeId` key.

### DIFF
--- a/linodecli/api_request.py
+++ b/linodecli/api_request.py
@@ -259,11 +259,13 @@ def _build_request_body(ctx, operation, parsed_args) -> Optional[str]:
     if ctx.defaults:
         parsed_args = ctx.config.update(parsed_args, operation.allowed_defaults)
 
+    param_names = {param.name for param in operation.params}
+
     expanded_json = {}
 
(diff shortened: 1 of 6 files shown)


ykim-akamai (approved):
LGTM, verified everything locally. Added simple integration test as well. Nice work!

zliang-akamai (approved):
LGTM, tests passed

PR #686, opened 2024-11-27
Replace lstrip with substring to resolve #685

### LINKED ISSUE #685: [Bug]: Lstrip is incorrect when removing bucket from prefix
### CLI Version

linode-cli 5.55.0 Built from spec version 4.191.1

### Command

linode-cli obj  put --acl-public js/buddy.js yb-cdn/buddy-dev

### Output

Uploading buddy.js:
 |############################################################| 100.0%
Done.

### Expected Behavior

The file is uploaded in bucket `yb-cdn` under prefix `buddy-dev`.

### Actual Behavior

The file that has been uploaded has not been uploaded under the prefix `buddy-dev/buddy.js` but under `uddy-dev/buddy.js`. It looks like the `lstrip` is the root cause of the issue -> `prefix = parsed.bucket.lstrip(f"{bucket}/")`.

The `lstrip` also matches with the `b` that was referenced in the `yb-cdn` which makes that the first `b` in `buddy` gets dropped.

### Steps to Reproduce

_No response_

jorenkoyen:
## 📝 Description

This PR fixes the issue described in #685.
It replaces the existing `lstrip` call to extract the object prefix with a substring based on the length of the bucket name.

## ✔️ How to Test

Tested full objects module and new test case which tries to reproduce the case described in #685.

* `make MODULE=obj testint`
* `make MODULE=obj TEST_CASE=test_obj_single_file_single_bucket_with_prefix_ltrim testint`

### DIFF
--- a/linodecli/plugins/obj/objects.py
+++ b/linodecli/plugins/obj/objects.py
@@ -91,7 +91,7 @@ def upload_object(
     bucket = parsed.bucket
     if "/" in parsed.bucket:
         bucket = parsed.bucket.split("/")[0]
-        prefix = parsed.bucket.lstrip(f"{bucket}/")
+        prefix = parsed.bucket[len(f"{bucket}/"):]
 
     upload_options = {
         "Bucket": bucket,
--- a/linodecli/plugins/obj/objects.py
+++ b/linodecli/plugins/obj/objects.py
@@ -91,7 +91,7 @@ def upload_object(
     bucket = parsed.bucket
     if "/" in parsed.bucket:
         bucket = parsed.bucket.split("/")[0]
-        prefix = parsed.bucket[len(f"{bucket}/"):]
+        prefix = parsed.bucket.removeprefix(f"{bucket}/")
 
     upload_options = {
         "Bucket": bucket,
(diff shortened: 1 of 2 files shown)


jorenkoyen:
I took the liberty of opening up a PR myself as it is only a small fix. I hope the integration test is up to spec as I'm not a day to day python developer.

yec-akamai (approved):
Thank you for putting up the fix!

linodecli/plugins/obj/objects.py
```diff
@@ -91,7 +91,7 @@ def upload_object(
     bucket = parsed.bucket
     if "/" in parsed.bucket:
         bucket = parsed.bucket.split("/")[0]
-        prefix = parsed.bucket.lstrip(f"{bucket}/")
+        prefix = parsed.bucket[len(f"{bucket}/"):]
```
zliang-akamai, line 94:
```suggestion
        prefix = parsed.bucket.removeprefix(f"{bucket}/")
```
Thanks for your contribution! I think `removeprefix` would do a better job.

jorenkoyen: Sure thing, not fully familiar with all the internal python string functions. I'll commit the changes to use `removeprefix`

zliang-akamai (approved):
Looks good to me!

zliang-akamai:
Thank you for your contribution! We will merge it after bumping the Python version because `removeprefix` requires 3.9.

PR #695, opened 2024-12-05
Only use +and in filters when necessary

lgarber-akamai:
## 📝 Description

This pull request updates the filter generation logic to only use `+and` when necessary, which should reduce the impact of bugs when filtering on special-case fields (e.g. `entity.type`, `entity.id`). 

## ✔️ How to Test

### Unit Testing

```
make testunit
```

### Manual Testing

1. Run the following command to create an instance with label `foobar` and tags `foobar` and `barfoo`:

```bash
linode-cli linodes create --type g6-nanode-1 --region us-mia --label foobar --tags foobar --tags barfoo
```

2. Run the following command to list instances matching the above conditions:

```bash
linode-cli linodes ls --label foobar --tags foobar --tags barfoo --debug
```

3. Ensure the debug output contains the following filter JSON:

```bash
X-Filter: {"label": "foobar", "+and": [{"tags": "foobar"}, {"tags": "barfoo"}]}
```

4. Ensure the instance is displayed in the command's output.

### DIFF
--- a/linodecli/api_request.py
+++ b/linodecli/api_request.py
@@ -162,26 +162,29 @@ def _build_filter_header(
     order_by = parsed_args_dict.pop("order_by")
     order = parsed_args_dict.pop("order") or "asc"
 
-    # The "+and" list to be used in the filter header
-    filter_list = []
+    result = {}
 
-    for k, v in parsed_args_dict.items():
-        if v is None:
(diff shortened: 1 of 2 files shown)


yec-akamai (approved):
Works well on my end!

jriddle-linode (approved):
LGTM and works locally

PR #698, opened 2024-12-09
Bump minimal Python version to 3.9

zliang-akamai:
## 📝 Description

Python 3.8 has been EOL, upgrade minimal Python version to 3.9 so we can use the features added in 3.9.


lgarber-akamai (approved):
Thank you for handling this!

ezilber-akamai (approved):

PR #699, opened 2024-12-09
fix: Updated OBJ plugin default cluster configurataion to be required

ezilber-akamai:
## 📝 Description

Requires that the user configures a default cluster when running `linode-cli obj mb` to prevent error.

## ✔️ How to Test

The following steps assume you have pulled down this PR locally and ran `make install`.

### Integration Tests
`make MODULE=obj testint`

### Unit Tests
`make testunit`

### Manual Tests
Run `linode-cli obj mb bug-test` and try not providing a default cluster to verify that the user is prompted for it again.

## 📷 Preview

image


zliang-akamai (approved):
Looks good and works well to me. Thanks for fixing it!

PR #705, opened 2024-12-17
Multiple workflow changes

zliang-akamai:
## 📝 Description

1. Re-organize workflows, combining multiple jobs into fewer workflows.
2. Add codeql and dependencies review jobs.
3. Remove unnecessary steps in different jobs.


.github/workflows/release.yml
```diff
@@ -1,9 +1,27 @@
-name: OCI Image Publish
+name: Release
 on:
   workflow_dispatch: null
   release:
     types: [ published ]
 jobs:
+  notify:
```
ykim-akamai, line 7: Can we move this step after all other published jobs and notify the channel if it succeeded?

ykim-akamai (approved):
LGTM, nice improvements

.github/workflows/release.yml
zliang-akamai: Thanks! Do you think `needs: pypi-release` is sufficient in this case? I think when docker publish failed we can manually re-release it.

jriddle-linode (approved):

PR #706, opened 2024-12-20
Align options

mssalvatore:
## 📝 Description

Configuration options are not aligned when presented to the user. In the below screenshot, you'll notice a shift in alignment when the option number increases from 9 to 10. A similar shift occurs when the numbers increase from 99 to 100. This PR aligns the options.

## ✔️ How to Test

Reproduce the issue by running `linode-cli configure`. Test the fix by running `make testunit`.


ezilber-akamai (approved):
Tested this out locally and it looks great. Thank you for the contribution!

ykim-akamai (approved):
LGTM, tested configuration steps locally and it looks great!

PR #715, opened 2025-01-06
fix: LKE integration test with node pool

yec-akamai:
## 📝 Description

The node pool related output is changed because of a recent doc fix. Update the integration test cases to reflect the change. 

## ✔️ How to Test

```
make MODULE=lke testint
```

https://github.com/linode/linode-cli/actions/runs/12639005395

### DIFF
--- a/tests/integration/lke/test_clusters.py
+++ b/tests/integration/lke/test_clusters.py
@@ -1,3 +1,5 @@
+import json
+
 import pytest
 
 from tests.integration.helpers import (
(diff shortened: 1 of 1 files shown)


lgarber-akamai (approved):
Tests are passing on my end, nice work!

yec-akamai:
The integration test passed in: https://github.com/linode/linode-cli/actions/runs/12639005395/job/35216438474

(The job failed on the Update PR check run, which is not related to the integration test itself.)

zliang-akamai (approved):

PR #723, opened 2025-02-03
Ssh by domain

jrouzierinverse:
## 📝 Description

Add the option to connect to a server using the domain to make it easier when connecting through the Akamai Zero Trust Client.

## ✔️ How to Test

make test-unit

Run the following

```linodes-cli ssh -d  label```

### DIFF
--- a/linodecli/plugins/ssh.py
+++ b/linodecli/plugins/ssh.py
@@ -50,6 +50,11 @@ def call(args, context):  # pylint: disable=too-many-branches
         action="store_true",
         help="If given, uses the Linode's SLAAC address for SSH.",
     )
+    parser.add_argument(
+        "-d",
+        action="store_true",
+        help="If given, uses the Lindoe's domain name for SSH",
+    )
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: linodecli/plugins/ssh.py [hunk 1]
```diff
@@ -147,4 +152,7 @@ def parse_target_address(
         if ip.startswith("192.168"):
             continue
 
+        if getattr(parsed, "d"):
+            ip = ip.replace(".", "-") + ".ip.linodeusercontent.com"
```
lgarber-akamai, line 156: Would it make sense to fetch the RDNS the instance's public IPv4 address(es) using the `linode-cli linodes ips-list --json` command to avoid hardcoding it?

jrouzierinverse: Make sense

lgarber-akamai (approved):
Works well in my local testing, thanks for the contribution!

zliang-akamai (approved):

Issue #735, opened 2025-02-27
[Bug]: Incomplete URL substring sanitization Unvalidated Redirects and Forwards


odaysec:
### CLI Version

v5.56.2

### Command

https://github.com/linode/linode-cli/blob/fafe73e1f48a48ab9cbdf9b01f679e041f6bf3fa/tests/integration/domains/test_slave_domains.py#L73-L73

Sanitizing untrusted URLs is a common technique for preventing attacks such as request forgeries and malicious redirections. Usually, this is done by checking that the host of a URL is in a set of allowed hosts. However, treating the URL as a string and checking if one of the allowed hosts is a substring of the URL is very prone to errors. Malicious URLs can bypass such security checks by embedding one of the allowed hosts in an unexpected location.

Even if the substring check is not used in a security-critical context, the incomplete check may still cause undesirable behaviors when the check succeeds accidentally.

## Recommendation
Parse a URL before performing a check on its host value, and ensure that the check handles arbitrary subdomain sequences correctly.

### Actual Behavior

CWE-20

## POC
The following code checks that a URL redirection will reach the `example.com` domain.
```python
from flask import Flask, request, redirect
from urllib.parse import urlparse

app = Flask(__name__)

# Not safe, as "evil-example.net/example.com" would be accepted

@app.route('/some/path/bad1')
def unsafe1(request):
    target = request.args.get('target', '')
    if "example.com" in target:
        return redirect(target)

# Not safe, as "benign-looking-prefix-example.com" would be accepted

@app.route('/some/path/bad2')
def unsafe2(request):
    target = request.args.get('target', '')
    if target.endswith("example.com"):
        return redirect(target)



#Simplest and safest approach is to use an allowlist

@app.route('/some/path/good1')
def safe1(request):
    allowlist = [
        "example.com/home",
        "example.com/login",
    ]
    target = request.args.get('target', '')
    if target in allowlist:
        return redirect(target)

#More complex example allowing sub-domains.

@app.route('/some/path/good2')
def safe2(request):
    target = request.args.get('target', '')
    host = urlparse(target).hostname
    #Note the '.' preceding example.com
    if host and host.endswith(".example.com"):
        return redirect(target)
```
[...]

PR #736, opened 2025-02-27
fix: Incomplete URL substring sanitization Unvalidated Redirects and Forwards

odaysec:
## 📝 Description
patched(fix) for #735

To fix the problem, we should parse the URL and check the hostname to ensure it matches the expected domain. This can be done using the `urlparse` function from the `urllib.parse` module. We will extract the hostname from the URL and check if it ends with the expected domain.

- Replace the substring check `"-example.com" in result` with a more robust check that parses the URL and verifies the hostname.
- Update the test to use the `urlparse` function to extract the hostname and check if it ends with the expected domain.

### DIFF
--- a/tests/integration/domains/test_slave_domains.py
+++ b/tests/integration/domains/test_slave_domains.py
@@ -70,7 +70,9 @@ def test_list_slave_domain(slave_domain):
     result = exec_test_command(
         BASE_CMD + ["list", "--text", "--no-header"]
     ).stdout.decode()
-    assert "-example.com" in result
+    from urllib.parse import urlparse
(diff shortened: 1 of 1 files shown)


ykim-akamai:
Thank you for updating the test case. Getting a minor assertion failure from `make test-int TEST_CASE=test_list_slave_domain`
```
AssertionError: assert (None)
E        +  where None = ParseResult(scheme='', netloc='', path='32190231741125163985209000-example.comslaveactive32190051741124805550044547exa...mple.clone-1740764427076302488-inttestsdk.orgmasteractivepathiel-test123@linode.com', params='', query='', fragment='').hostname
```

PR #737, opened 2025-02-27
Print OBJ plugin errors to stderr

ezilber-akamai:
## 📝 Description

Updated `boto3` errors in the OBJ plugin to print to stderr instead of stdout.


ykim-akamai (approved):
LGTM, nice work! Should we create a test ticket to write some negative test cases?

ezilber-akamai:
> LGTM, nice work! Should we create a test ticket to write some negative test cases?

Good idea!

jriddle-linode (approved):
LGTM!

PR #747, opened 2025-03-14
Add conditions to notification job in e2e workflow

ykim-akamai:
## 📝 Description

` needs: [integration_tests, process-upload-report]`
The condition above was missed which was causing the summary to be empty during notification job

## ✔️ How to Test

Tested on forked and private channel 
image

## 📷 Preview


lgarber-akamai (approved):
Thanks for fixing this!

PR #755, opened 2025-04-02
Add support for custom aliases for commands

ezilber-akamai:
## 📝 Description

Users can now add a custom alias for a command by using:
`linode-cli set-custom-alias --command linodes --alias lins`

and remove a custom alias for a command by using:
`linode-cli set-custom-alias --command linodes --alias lins`

## 📷 Preview

image

### DIFF
--- a/linodecli/__init__.py
+++ b/linodecli/__init__.py
@@ -110,6 +110,53 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
         # if not spec was found and we weren't baking, we're doomed
         sys.exit(ExitCodes.ARGUMENT_ERROR)
 
+    if parsed.command == "set-custom-alias" or parsed.command == "remove-custom-alias":
+        # Ensure both --command and --alias are provided
+        # TODO: Can we use argparse for these
+        if len(args) == 4:
+            if args[0] != "--command" or args[2] != "--alias":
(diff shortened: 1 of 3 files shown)


linodecli/__init__.py
```diff
@@ -110,6 +107,69 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
         # if not spec was found and we weren't baking, we're doomed
         sys.exit(ExitCodes.ARGUMENT_ERROR)
 
+    if parsed.command in ("set-custom-alias", "remove-custom-alias"):
+        # Ensure both --command and --alias are provided
+        # TODO: Can we use argparse for these
+        if len(args) == 4:
+            if args[0] != "--command" or args[2] != "--alias":
```
zliang-akamai, line 114: Do you think we can add `"command"` and `"alias"` into the Python argument parser to avoid manually parsing them here? I think we will be able to support out of order set alias command by that way, like `linode-cli set-custom-alias --alias [ALIAS] --command [COMMAND]`

ezilber-akamai: Yes, good idea! I updated this, but I had to switch `--command` to `--target-command` since `command` is already an argument in the argument parser.

zliang-akamai (approved):
Looks good and works well for me. Nice work!

ykim-akamai (approved):
LGTM, nice work!

PR #757, opened 2025-04-07
Error on missing x-linode-cli-command extension; log when operationId is used for action

lgarber-akamai:
## 📝 Description

This pull request adds logic to raise an error if an OpenAPI path has CLI-enabled actions but does _not_ specify an `x-linode-cli-command` extension. This is necessary because we've recently had OpenAPI specification release candidates make it very far down the pipeline before commands misgrouped under `default` are discovered, which is not intentionally used by any existing commands. 

This pull request also adds a new log entry that occurs when the `x-linode-cli-action` extension is not specified and the operationId is used as a failover.

## ✔️ How to Test

### Unit Testing

```
make test-unit
```

### DIFF
--- a/linodecli/cli.py
+++ b/linodecli/cli.py
@@ -73,7 +73,12 @@ class CLI:  # pylint: disable=too-many-instance-attributes
         }
 
         for path in spec.paths.values():
-            command = path.extensions.get(ext["command"], "default")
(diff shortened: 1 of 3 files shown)


yec-akamai (approved):
Works well on my end!

ezilber-akamai (approved):
LGTM!

PR #762, opened 2025-04-26
Rename workflows; cleanup unnecessary steps

zliang-akamai:
## 📝 Description
This is to cleanup the workflows and rename them to reflect the platform they will run on.


yec-akamai (approved):
LGTM!

PR #763, opened 2025-04-26
Trusted publisher for PyPI

zliang-akamai:
## 📝 Description

Migrate to trusted publisher to avoid token usage.

Sample release: https://github.com/zliang-akamai/linode-cli/actions/runs/14677674503/job/41196302382

Do we want to configure [some protection rules for the `pypi-release` environment](https://github.com/linode/linode-cli/settings/environments/6488209563/edit)?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -79,7 +79,11 @@ jobs:
             github_token=${{ secrets.GITHUB_TOKEN }}
 
   pypi-release:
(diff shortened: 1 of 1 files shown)


lgarber-akamai (approved):
Looks great! Always nice to reduce the number of secrets we have to keep our eyes on 🙂

yec-akamai (approved):

PR #771, opened 2025-05-07
Update disk encryption node pool test

ykim-akamai:
## 📝 Description

Test is made more flexible to check if disk encryption value is set instead of seeing the default value always equal to disabled

## ✔️ How to Test

`make test-int TEST_CASE=test_create_node_pool_has_disk_encryption_field `

## 📷 Preview


yec-akamai (approved):
Thanks for fixing it!

ezilber-akamai (approved):
Thanks for the fix!

PR #776, opened 2025-05-23
Added override for DB Engine Config view commands

ezilber-akamai:
## 📝 Description

Added override to properly construct tables for DB Engine Config view commands.

## ✔️ How to Test

Manually run `linode-cli databases mysql-config-view` and `linode-cli databases postgres-config-view` after checking out this PR and building against the proper spec. Confirm that the tables are properly populated and look correct.


lgarber-akamai (approved):
Looks great and works well on my end!

PR #792, opened 2025-07-07
Fix usage of `yaml.load` in getkubeconfig plugin

ezilber-akamai:
## 📝 Description

Replace usage of `yaml.load` with `yaml.safe_load` in get-kubeconfig plugin.

## ✔️ How to Test

`make test-unit`


lgarber-akamai (approved):
Thanks for the quick fix!

PR #805, opened 2025-08-07
Fixed broken integration test

ezilber-akamai:
## 📝 Description

Fixed integration test that was failing due to spec change.

## ✔️ How to Test

`make test-int TEST_CASE=test_view_pool`
`make test-int TEST_CASE=test_node_pool_list`


lgarber-akamai (approved):
Nice work on digging into this!

zliang-akamai (approved):

PR #807, opened 2025-08-08
Pass JSON value parameter as is to the API call

zliang-akamai:
## 📝 Description

Previously the value of a JSON parameter passed in by the user had to go through `_traverse_request_body` function to have null or empty list/dict values remove. We did have a way to allow explicit empty dict/list but that won't work if the empty dict/list is nested in the JSON object.

Here is an example:
```bash
linode-cli linodes create \
  --image 'linode/ubuntu24.04' \
  --interface_generation linode \
  --interfaces '[{"firewall_id":123456,"public":{}}]' \
  --private_ip false \
  --region us-mia \
  --type g6-nanode-1 \
  --label test-cli-create-linode-public \
  --root_pass 'this-is-an-unsafe-password' \
  --disk_encryption enabled --debug
```

The JSON value parameter `interfaces` with user input JSON string `'[{"firewall_id":123456,"public":{}}]'` would be translated into an invalid API call request body attribute, which had `{}` removed:

```json
{
    "interfaces": [{"firewall_id": 2431805}]
}
```
(irrelevant attributes in the body are ignored)

This change will let the CLI to pass whatever user input as-is into the request body to resolve this issue.

This is a potential breaking change because if user passed in some JSON like `{"a": {}, b: {"key": "value"}}` and expected key `a` would be dropped, this change will break their code. But I believe this change is necessary to make CLI behaviors correct and there isn't another easy way to fix the bug.

## ✔️ How to Test
```bash
make install && linode-cli linodes create \
  --image 'linode/ubuntu24.04' \
  --interface_generation linode \
  --interfaces '[{"firewall_id":1629606,"public":{}}]' \
  --private_ip false \
  --region us-ord \
  --type g6-nanode-1 \
  --label test-cli-create-linode-public \
  --root_pass 'this-is-an-unsafe-password' \
  --disk_encryption enabled --debug
```

### DIFF
--- a/linodecli/api_request.py
+++ b/linodecli/api_request.py
@@ -20,6 +20,7 @@ from linodecli.helpers import API_CA_PATH, API_VERSION_OVERRIDE
 from .baked.operation import (
     ExplicitEmptyDictValue,
     ExplicitEmptyListValue,
+    ExplicitJsonValue,
     ExplicitNullValue,
     OpenAPIOperation,
 )
@@ -320,6 +321,10 @@ def _traverse_request_body(o: Any) -> Any:
 
             if isinstance(v, ExplicitNullValue):
                 result[k] = None
+                continue
+
+            if isinstance(v, ExplicitJsonValue):
+                result[k] = v.json_value
                 continue
 
             value = _traverse_request_body(v)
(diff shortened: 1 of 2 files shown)


ezilber-akamai (approved):
Nice work!

PR #809, opened 2025-08-09
Fix minor grammar error

FelicianoTech:
When the text is outputted to terminal, there's no space between the words. Based on other strings, it seems like the convention is to add the space to the end of a string.


zliang-akamai (approved):
Thanks for your contribution!

yec-akamai (approved):
Thanks for the fix!

PR #857, opened 2026-02-20
Support x-linode-cli-skip at attribute level

dawiddzhafarov:
## 📝 Description

This PR adds support for `x-cli-linode-skip` extension on attribute level in requests and responses.

## ✔️ How to Test

Run unit tests:
`make test-unit`

You can also modify openAPI spec and add `x-cli-linode-skip` to the desired attributes and bake the CLI. Use the command and verify manually that the attribute(s) are not present.


vshanthe (approved):

yec-akamai (approved):
Works well on my end!

PR #878, opened 2026-04-04
feat: support "help" as a subcommand synonymous with "--help" #868

### LINKED ISSUE #868: [Feature]: support "help" as a subcommand synonymous with "--help"
### Description

Across all the different cloud providers, it becomes easy to forget which option brings up the documentation, manual page, or usage.  I frequently find myself typing `linode <command> help`, only to then have to go back and running `linode <command> --help` (or `-h`).

It'd be nice if `help` as a subcommand was synonymous with `--help`.  It'd be even nicer if a `help` subcommand brought up a full manual page similar to how AWS cli tools do it.  Ideally, you'd then have `linode <command> help` get you the full, detailed manual page in the default pager and `lincode <command> --help` give you the summary of supported command-line options on stdout only.

### Potential Usage Example

_No response_

dhruveshmishra:
Added support for using `help` as a subcommand in the CLI.

Closes #868 

What this does
Now the following commands work:

linode-cli instance help
linode-cli help instance

Both behave the same as using `--help`.

Why
While using different CLIs, I often find myself typing `help` as a subcommand instead of `--help`. Many tools like git and aws support this pattern, so adding it here makes the CLI a bit more intuitive and user-friendly.

Implementation
This is handled by checking for `help` in the arguments early and mapping it to `--help`. The change is minimal and doesn’t affect existing functionality.

Notes
If there’s a preferred way to handle this at the parser level instead of rewriting argv, I’m happy to update the implementation.

### DIFF
--- a/linodecli/cli.py
+++ b/linodecli/cli.py
@@ -25,7 +25,13 @@ from linodecli.output.output_handler import OutputHandler, OutputMode
 METHODS = ("get", "post", "put", "delete")
 
 logger = getLogger(__name__)
+args = sys.argv[1:]
 
+if len(args) > 0 and args[-1] == "help":
+    sys.argv = [sys.argv[0]] + args[:-1] + ["--help"]
+
+elif len(args) > 0 and args[0] == "help":
+    sys.argv = [sys.argv[0]] + args[1:] + ["--help"]
 
 class CLI:  # pylint: disable=too-many-instance-attributes
     """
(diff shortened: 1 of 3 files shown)


mgwoj (changes requested):
Please follow the CoPilot suggestions

dhruveshmishra:
@mgwoj  there are a little bit changes from co- pilot which ai gives unnessesary for the expansion of code the solution i had done covered the main portion which fix the issue so please consider it as a corrected code

PR #886, opened 2026-05-08
TPT-4280: Add integration tests for Reserved IP for IPv4

mawilk90:
## 📝 Description

Integration tests verifying Reserved IPs CLI commands

NOTE:
[PR](https://git.source.akamai.com/projects/TECHDOCS/repos/akamai-api-inventory/pull-requests/2707/overview) with openapi.json for Reserved IPs is already merged, however openapi.json file is not updated yet in `linode-api-openapi` repo so CLI can be build locally only using the latest openapi.json taken from the PR mentioned above

## ✔️ How to Test

`make test-int TEST_SUITE=networking TEST_ARGS="-k reserved"`
`make test-int TEST_SUITE=tags TEST_ARGS="-k reserved"`

### DIFF
--- a/tests/integration/linodes/helpers.py
+++ b/tests/integration/linodes/helpers.py
@@ -1,4 +1,5 @@
 import json
+import os
 import time
 
 from tests.integration.helpers import (
@@ -7,7 +8,7 @@ from tests.integration.helpers import (
 )
 
 DEFAULT_RANDOM_PASS = exec_test_command(["openssl", "rand", "-base64", "32"])
-DEFAULT_REGION = "us-ord"
+DEFAULT_REGION = "pl-labkrk-2" if "devcloud" in os.getenv('LINODE_CLI_API_HOST') else "us-ord"
 
 DEFAULT_TEST_IMAGE = exec_test_command(
     [
--- a/tests/integration/linodes/helpers.py
+++ b/tests/integration/linodes/helpers.py
@@ -8,7 +8,11 @@ from tests.integration.helpers import (
 )
 
 DEFAULT_RANDOM_PASS = exec_test_command(["openssl", "rand", "-base64", "32"])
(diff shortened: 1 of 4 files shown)


tests/integration/linodes/helpers.py
```diff
@@ -7,7 +8,11 @@
 )
 
 DEFAULT_RANDOM_PASS = exec_test_command(["openssl", "rand", "-base64", "32"])
-DEFAULT_REGION = "us-ord"
+DEFAULT_REGION = (
+    "pl-labkrk-2"
+    if "devcloud" in os.getenv("LINODE_CLI_API_HOST")
```
ezilber-akamai:
```suggestion
    if "devcloud" in os.getenv("LINODE_CLI_API_HOST", "")
```

When I ran the tests, they initially errored out because I had `LINODE_CLI_API_HOST` unspecified. Adding a default here should fix that so that we can let it properly default to the prod environment.

ezilber-akamai (commented):
Tests look good and pass locally! One small issue though, left a comment about it.

tests/integration/networking/fixtures.py
```diff
@@ -0,0 +1,70 @@
...
+
+    result = json.loads(exec_test_command(command))[0]
+
+    yield result
+
+    delete_target_id("networking", result["address"], "reserved-ip-delete")
+
+
+@pytest.fixture(scope="package")
+def test_linode_id(linode_cloud_firewall):
```
psnoch-akamai: How about renaming this function? It starts with test, what suggest that is is test case

psnoch-akamai (approved):
LGTM, tests work locally

tests/integration/networking/fixtures.py
mawilk90: Good catch! I missed that moving existing fixtures to networking. I've already renamed them to avoid this kind of ambiguity

ezilber-akamai (approved):

PR #887, opened 2026-05-11
TPT-4264: Create tests for Create Linodes Options and Password-less

psnoch-akamai:
## ✔️ How to Test

make test-int TEST_CASE=test_expected_error_if_fields_authorized_users_authorized_keys_root_pass_are_not_set
make test-int TEST_CASE=test_create_linode_with_kernel_and_boot_size_then_add_disk_and_rebuild

TECH doc PR: https://git.source.akamai.com/projects/TECHDOCS/repos/akamai-api-inventory/pull-requests/2685/overview - to build it fix from https://github.com/linode/linode-cli/pull/888 is needed

### DIFF
--- a/tests/integration/linodes/helpers.py
+++ b/tests/integration/linodes/helpers.py
@@ -312,19 +312,25 @@ def wait_for_disk_status(
     must_end = time.time() + timeout
     while time.time() < must_end:
         time.sleep(period)
-        result = exec_test_command(
-            [
-                "linode-cli",
-                "linodes",
-                "disk-view",
-                linode_id,
-                disk_id,
-                "--format",
-                "status",
-                "--text",
-                "--no-headers",
-            ]
-        )
+        try:
+            result = exec_test_command(
+                [
+                    "linode-cli",
+                    "linodes",
+                    "disk-view",
+                    linode_id,
+                    disk_id,
+                    "--format",
+                    "status",
+                    "--text",
+                    "--no-headers",
+                ]
+            )
+        except RuntimeError as response_error:
+            if "Not found" in str(response_error):
+                continue
+            else:
+                raise RuntimeError(response_error)
         if status == result:
             return True
     return False
--- a/tests/integration/linodes/helpers.py
+++ b/tests/integration/linodes/helpers.py
@@ -304,3 +304,27 @@ def get_disk_id(test_linode_instance):
     ).splitlines()
     first_id = disk_id[0].split(",")[0]
     return first_id
+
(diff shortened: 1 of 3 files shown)


tests/integration/linodes/helpers.py
```diff
@@ -304,3 +304,27 @@ def get_disk_id(test_linode_instance):
...
     return first_id
+
+
+def wait_for_disk_status(
+    linode_id: "str", disk_id: "str", timeout, status: "str", period=10
+):
+    must_end = time.time() + timeout
+    while time.time() < must_end:
+        time.sleep(period)
+        result = exec_test_command(
```
psnoch-akamai: Please verify if test pass for you. I get error randomly:

> E           Request failed: 404
E           	Not found

mawilk90:
It's worth to mention that using openapi.json from TECH doc PR mentioned in the description will fail with the current linode-cli's parser. Solution provided [here](https://github.com/linode/linode-cli/pull/888)

tests/integration/linodes/helpers.py
```diff
@@ -304,3 +304,27 @@ def get_disk_id(test_linode_instance):
     ).splitlines()
     first_id = disk_id[0].split(",")[0]
     return first_id
+
+
+def wait_for_disk_status(
+    linode_id: "str", disk_id: "str", timeout, status: "str", period=10
```
mawilk90, line 310: Is it not possible to use existing `wait_for_condition` instead of creating a new method?

tests/integration/linodes/helpers.py
mawilk90: Yes, it fails so maybe usage of `wait_for_condition` is going to help (as mentioned [here](https://github.com/linode/linode-cli/pull/887#discussion_r3225863766))?

ezilber-akamai: I'm also getting the 404

tests/integration/linodes/helpers.py
psnoch-akamai: Good idea, but probably I need to use custom one, because of 404 error

tests/integration/linodes/helpers.py
psnoch-akamai: The same result with wait_for_condition. Seems like API issue and if status is not ready it is not available for some time

ezilber-akamai (approved):
Tests passing locally. Nice work!

mawilk90 (approved):

PR #911, opened 2026-07-29
TPT-4560: Add merge readiness check

mawilk90:
## 📝 Description

Added extra step in `ci.yml` workflow to verify if the PR is ready for merging (i.e. it doesn't have a `do-not-merge` label) and fail the workflow otherwise.

Notes:
- the workflow activates when PR labels are modified
- the workflow depends on `do-not-merge` label presence (it fails when the label has been added)

## ✔️ How to Test

Please play with PR's labels on the right and check Labeler workflow status below

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,7 @@ on:
   workflow_dispatch: null
   push:
   pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
 jobs:
   docker-build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


psnoch-akamai (approved):
LGTM. Just after merge, we can probably find ready for merge PR, add label and test if lock works.

This one looks like good candidate https://github.com/linode/linode-cli/pull/875

lgarber-akamai (approved):
Great work!

.github/workflows/ci.yml
```diff
@@ -58,6 +59,18 @@ jobs:
...
+        run: |
+          LABELS='${{ toJson(github.event.pull_request.labels.*.name) }}'
+
+          if echo "$LABELS" | grep -q "do-not-merge"; then
+            echo "This PR is labeled as 'do-not-merge' so please make sure it is ready."
+            exit 1
+          else
+            exit 0
+          fi
+
```
zliang-akamai: I think it would be cleaner and safer (against [script injection](https://securitylab.github.com/resources/github-actions-untrusted-input/)) to put both conditions in the `if`

```suggestion
      - name: Check Merge Readiness
        if: github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'do-not-merge')
        run: |
          echo "::error::This PR is labeled 'do-not-merge'."
          exit 1
```

PR #919, opened 2026-08-07
TPT-4221: Enhance security of kubeconfig file permission

yec-akamai:
## 📝 Description

Ensure that kubeconfig files are now created via 600 instead of 644. Permissions are set atomically at creation, so contents are never briefly world-readable. Pre-existing files with group/other bits set are tightened to 600 on write. The parent directory is created with mode 700.

## ✔️ How to Test

```
make test-unit
```

### DIFF
--- a/linodecli/plugins/get-kubeconfig.py
+++ b/linodecli/plugins/get-kubeconfig.py
@@ -8,6 +8,7 @@ Usage:
 
 import argparse
 import base64
+import os
 import sys
 from pathlib import Path
 
@@ -18,6 +19,11 @@ from linodecli.help_formatter import SortingHelpFormatter
 from linodecli.plugins import inherit_plugin_args
 
 PLUGIN_BASE = "linode-cli get-kubeconfig"
+
+# Kubeconfigs contain credentials, so they should only be
+# accessible by the user that created them.
+KUBECONFIG_FILE_MODE = 0o600
+KUBECONFIG_DIR_MODE = 0o700
 
 
 def call(args, context):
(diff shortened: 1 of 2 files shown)


mawilk90 (approved):

linodecli/plugins/get-kubeconfig.py
```diff
@@ -147,9 +153,26 @@ def _load_config(filepath):
...
+
+    # Tighten the permissions of pre-existing files that are readable or
+    # writable by users other than the owner.
+    # NOTE: os.fchmod is not available on Windows, where POSIX file modes
+    # are not meaningful anyway.
+    if hasattr(os, "fchmod") and os.fstat(file_descriptor).st_mode & 0o077:
+        os.fchmod(file_descriptor, KUBECONFIG_FILE_MODE)
+
+    with os.fdopen(file_descriptor, "w", encoding="utf-8") as file:
+        yaml.dump(data, file)
```
zliang-akamai:
```suggestion
    filepath.parent.mkdir(mode=KUBECONFIG_DIR_MODE, exist_ok=True)

    def opener(path, flags):
        return os.open(path, flags, mode=KUBECONFIG_FILE_MODE)

    with open(filepath, "w", encoding="utf-8", opener=opener) as file_descriptor:
        # Tighten the permissions of pre-existing files that are readable or
        # writable by users other than the owner.
        # NOTE: os.fchmod is not available on Windows, where POSIX file modes
        # are not meaningful anyway.
        if hasattr(os, "fchmod"):
            os.fchmod(file_descriptor.fileno(), KUBECONFIG_FILE_MODE)

        yaml.dump(data, file_descriptor)
```

I think we can make it simpler this way

zliang-akamai (approved):

PR #921, opened 2026-08-27
Using official naming conventions; use only login-action for OIDC

zliang-akamai:
- Using docker official naming conventions for var names
- Use only one action for logging int via OIDC


mawilk90 (approved):

lgarber-akamai (approved):
Looks great, nice work!

